PR #79, opened 2022-11-24
feat: cid.contact

scott-dn:
1. indexer request.
2. simple cache and round-robin selection for the list of nodes.


scott-dn:
handle in #107

PR #81, opened 2022-11-28
feat: global metrics + http tracker

ozwaldorf:
- simple http tracker
  - provides prometheus http service discovery
  - ssl
- nodes register config via tracker (default set to `tracker.ursa.earth`)
- detailed metrics mirrored from libp2p_metrics ([example](https://kademlia-exporter.max-inden.de/d/Pfr0Fj6Mk/rust-libp2p))
  - swarm
  - ping
  - identify
  - kad
  - gossipsub
  - request response
- detailed metrics exposed from bitswap
- terraform tracker instance additions. via docker compose:
  - nginx + letsencrypt
  - prometheus (https://prometheus.ursa.earth)
  - grafana (https://grafana.ursa.earth)
  - http tracker (https://tracker.ursa.earth)
- refactor to `Recorder` trait implementations for each event type
- refactor metrics http handlers
- refactor rpc middleware
- refactored http/rpc server to include metrics endpoints under a single server/port
- removed metrics config (all http is under one port now, and all metrics are found at `/metrics`)

Which allows us to build nice grafana views with nodes being auto-discovered/scraped:
image
image


infra/tf/main.tf
```diff
@@ -71,42 +70,37 @@ resource "digitalocean_droplet" "bootstrap-node" {
...
-  user_data  = file(format("%s/bootstrap_node.yml", path.module))
+  user_data  = file("${path.module}/configs/bootstrap_node.conf")
   ssh_keys = [
     data.digitalocean_ssh_key.ursa-dev.id
   ]
 }
 
 resource "digitalocean_firewall" "ursa-network" {
-  count = var.node_count
+  count = 1
```
b0xtch, line 80:
```suggestion
  count = var.node_count
```

infra/tf/main.tf
```diff
@@ -71,42 +70,37 @@ resource "digitalocean_droplet" "bootstrap-node" {
...
   name  = "ursa-network-only-allow-ssh-http-and-https"
-
   droplet_ids = concat(digitalocean_droplet.testnet-node[*].id, digitalocean_droplet.bootstrap-node[*].id)
 
   inbound_rule {
     protocol         = "tcp"
     port_range       = "22"
     source_addresses = ["0.0.0.0/0", "::/0"]
   }
-
```
b0xtch, line 91: comment on these red changes

crates/ursa-network/src/service.rs
```diff
@@ -236,6 +240,45 @@ where
             event_receiver,
             response_channels: Default::default(),
             index_provider,
+            config: config.clone(),
+        }
+    }
+
+    // Make an http node announcement to a ursa tracker. Used for mapping the network pre-consensus
+    pub async fn announce_node(&self) -> Result<String> {
```
b0xtch, line 248: Perhaps using a another name than `announce_node` too similar to announcing in index-provider

infra/tf/terraform.tfstate
```diff
@@ -1,7 +1,7 @@
 {
```
b0xtch, line 1: we can remove this file and the `tf.state.backup`

crates/ursa-tracker/self_signed_certs/cert.pem
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
b0xtch, line 1: this file and `key.pem` should not be committed

infra/tf/configs/grafana/cloud.conf
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
b0xtch, line 1: could we instead put these into a docker-compose just like we do with each node. Including the tracker, we just `dcoker-compose up`

b0xtch (approved):
Nice work, could we perhaps also add store metrics to each node. Also maybe a TTFB for the globally telemetry.

crates/ursa-network/src/config.rs
```diff
@@ -34,6 +34,11 @@ pub struct NetworkConfig {
     pub identity: String,
     /// Keystore path. Defaults to ~/.ursa/keystore
     pub keystore_path: PathBuf,
+    /// Temporary HTTP tracker url. This is used for pre-consensus node announcements.
+    /// Defaults to devnet tracker.
+    pub tracker: String,
+    /// Prometheus metrics port
+    pub metrics_port: Option<u16>,
```
theBeardA, line 41: If a user needs to change the metrics port, they would need to update in 2 places in config file?

infra/tf/main.tf
ozwaldorf: So the reason I set this to 1 is:
- the existing DO deployment only had a single firewall
- a few lines down, for droplet_ids, it selects all bootstrap and testnet nodes. If we want an individual firewall for each, it should be configured as such, otherwise we end up with 7 firewalls on every node.

crates/ursa-network/src/service.rs
ozwaldorf: Agreed its a little confusing on the terminology. 

Maybe `register_node` might be better?

infra/tf/terraform.tfstate
ozwaldorf: will remove all extraneous files :+1:

infra/tf/main.tf
ozwaldorf: Styling consistency (we don't use double newlines anywhere else in the terraform code)

b0xtch:
Yeah it would be from outside

crates/ursa-network/src/service.rs
ozwaldorf: 2c8ebc9a0959fad47629f6efb80f6b3898261123

infra/tf/configs/grafana/cloud.conf
ozwaldorf: With this are you saying that grafana/prometheus/tracker should all be ran on a single machine with a docker compose?

crates/ursa-tracker/src/types.rs
```diff
@@ -0,0 +1,79 @@
+use hyper::{client::HttpConnector, Body};
+use libp2p::PeerId;
+use serde_derive::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+pub type Client = hyper::client::Client<HttpConnector, Body>;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NodeAnnouncement {
```
b0xtch, line 9: Just a quick reminder on the naming here conflicting with index announcement

ozwaldorf: done :smiley:

crates/ursa-tracker/src/main.rs
```diff
@@ -0,0 +1,228 @@
...
+        let db = db();
+        let id = PeerId::random();
+
+        let res = make_announcement(db.clone(), Some("8.8.8.8".to_string()), id).await;
+        info!("{:?}", res);
+        assert_eq!(res.0, 200);
+        db.remove(id.to_string().as_bytes()).unwrap();
+    }
+
+    // doesn't work with ipinfo
```
b0xtch, line 193: Dead code

infra/tf/terraform.tfstate
b0xtch: Reminder

infra/tf/main.tf
```diff
@@ -71,42 +70,41 @@ resource "digitalocean_droplet" "bootstrap-node" {
   size       = var.droplet_size
   backups    = false
   monitoring = true
-  user_data  = file(format("%s/bootstrap_node.yml", path.module))
+  user_data  = file("${path.module}/configs/bootstrap_node.yml")
```
b0xtch, line 73: Any benefit to this change?

ozwaldorf: cleaner code, and we're not doing any special formatting to really warrant using the full format when just using variable replacement works fine

crates/ursa-tracker/src/main.rs
ozwaldorf: Actually need to restore this functionality, will push soon

infra/tf/main.tf
b0xtch: I don’t agree that it’s cleaner code

b0xtch: It’s more cognitive load to figure out why the change was made vs subjective cleaner code change

b0xtch: Let’s try to go over the PR and revert such subjective cleaner code changes

crates/ursa-tracker/src/main.rs
b0xtch: Do not dead code? Or did you forget to push it?

ozwaldorf: When i refactored the ip info api from `ip-api.com` -> `ipinfo.io`, the functionality for fetching dns was lost, glad you caught it as I still needed to provide an alternative for the new api

infra/tf/main.tf
ozwaldorf: Okay sure. Although the reason was to also follow the same coding style with line 158 in the same file; which would be hard to read with the additional function call vs inline variable. I can still revert and modify the other line if you think it's better

infra/tf/terraform.tfstate
ozwaldorf: I removed with https://github.com/fleek-network/ursa/pull/81/commits/fb2940feddf3c36100b10341ace27fe7f07ac7da

But later did a `terraform plan` realized without that file, terraform cannot correctly create a plan and will attempt to create the entire network from scratch and readded with https://github.com/fleek-network/ursa/pull/81/commits/9a28c59e07427354edb855d22c88847a01623575

crates/ursa-tracker/src/main.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/81/commits/e639fe86f04b9d57cca0c7748673ea79689f4332

infra/tf/terraform.tfstate
ozwaldorf: removed again: https://github.com/fleek-network/ursa/pull/81/commits/bb8a4e5c689c16e6608277422a0dac68238c1a43

crates/ursa-metrics/Cargo.toml
```diff
@@ -2,18 +2,30 @@
 name = "ursa-metrics"
 version = "0.1.0"
 edition = "2021"
-authors = ["mtavano <marcello@fleek.co>"]
+authors = ["ossian <oz@fleek.xyz>", "mtavano <marcello@fleek.co>"]
```
b0xtch, line 5: i would use the fleek.network email here

b0xtch:
Just testing the urls above, getting 502 bad gateway. SSL certs don't seem to be valid

ozwaldorf:
> Just testing the urls above, getting 502 bad gateway. SSL certs don't seem to be valid

Was redeploying and ran to the store, it got stuck because the cloud config was set to use main (preperation for merging). Just manually checked out via ssh and reran the commands. Everything is back up now

infra/tf/main.tf
```diff
@@ -117,15 +115,90 @@ resource "digitalocean_firewall" "ursa-network" {
...
+}
+
+resource "digitalocean_record" "prometheus-domain" {
+  domain = digitalocean_domain.default.name
+  type   = "A"
+  name   = "prometheus"
+  value  = digitalocean_droplet.http-tracker.ipv4_address
+}
+
+resource "digitalocean_droplet" "http-tracker" {
```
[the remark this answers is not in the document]
ozwaldorf: yes https://github.com/fleek-network/ursa/pull/81/commits/b49560436b5195157e0151065fafdb68d63880c9

crates/ursa-tracker/src/ip_api.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    if !status.is_success() {
+        return Err(anyhow!("Failed to get ip info: {}", status));
+    }
+
+    let mut info: IpInfoResponse = serde_json::from_slice(&data)?;
+    let loc = info.loc.split(',').collect::<Vec<&str>>();
+    let lat = loc[0].parse::<f64>()?;
+    let lon = loc[1].parse::<f64>()?;
+    info.geo = geohash(lat, lon)?;
```
qti3e, line 59: why are we changing the response based on the response? what does ipinfo.io return on the `.geo` part?

crates/ursa-tracker/src/ip_api.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub country: String,
+    pub loc: String,
+    pub org: String,
+    pub postal: String,
+    pub timezone: String,
+    pub geo: String,
+}
+
+/// Get public ip info from https://ipinfo.io
+pub async fn get_ip_info(token: String, addr: String) -> Result<IpInfoResponse> {
```
qti3e, line 28: I think it would be nicer if the `ipinfo.io`'s type was not publicly exposed as the service which is used should be an implementation detail, and we had an owned struct which we could rely on its format without the underlying service.

crates/ursa-tracker/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+    headers: HeaderMap,
+    ConnectInfo(req_addr): ConnectInfo<SocketAddr>,
+    db: Extension<Arc<DB>>,
+    token: Extension<String>,
+    Json(registration): Json<TrackerRegistration>,
+) -> (StatusCode, Json<Value>) {
+    let id = registration.id;
+    info!("Received registration for: {}", id);
+
+    // todo: registration verification
```
qti3e, line 59: in case the todo is forgotten.

crates/ursa-tracker/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    let addr = registration.addr.clone().unwrap_or_else(|| {
+        // if no dns or ip address is provided, use the address of the request.
+        // Prefer X-Forwarded-For header if present from reverse proxy. otherwise, use the
+        // address of the request
+        let ip = headers
+            .get("X-Forwarded-For")
+            .map(|x| x.to_str().unwrap().to_string())
+            .unwrap_or_else(|| req_addr.ip().to_string());
+        (ip != "127.0.0.1").then_some(ip).unwrap_or_default()
```
qti3e, line 69: this code possibly allows an attacker to cause the termination of the server by putting non-visible ASCII characters in the `X-Forwarded-For` header in scenarios where there is no reverse-proxy overwrite of the said header.

this is caused by the `x.to_str().unwrap()` part of the code.

crates/ursa/src/main.rs
```diff
@@ -64,6 +66,24 @@ async fn main() {
 
                 let keypair = im.current();
 
+                let registration = TrackerRegistration {
+                    id: keypair.clone().public().to_peer_id(),
+                    // TODO: calculate or get from config the supplied storage in bytes
```
qti3e, line 71: in case it's forgotten.

qti3e (commented):
looks good! i just left a few comments.

crates/ursa-tracker/src/ip_api.rs
ozwaldorf: ipinfo.io (and others) doesn't provide geohash unfortunately. But it can be hashed from lat and long easily

crates/ursa/src/main.rs
ozwaldorf: Not sure on how exactly storage should be computed. I think the ursa binary needs this feature first, then whatever that uses can pass a value here

crates/ursa-metrics/src/request_response.rs
```diff
@@ -0,0 +1,45 @@
...
+                        );
+                    }
+                    RequestResponseMessage::Response { request_id, .. } => {
+                        increment_counter!("req-res_total_response_sent");
+                        increment_counter!(
+                            "req-res_response_sent",
+                            vec![
+                                Label::new("peer", format!("{}", peer)),
+                                Label::new("request_id", request_id.to_string()),
+                                // channel?
```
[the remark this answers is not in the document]
ozwaldorf: This was implemented in previous metrics but I didn't think it was necessary to include it

crates/ursa-tracker/src/main.rs
ozwaldorf: Not sure what it'd entail/not required at this time. cc @b0xtch but basically once we have consensus we can switch to that

crates/ursa-tracker/src/ip_api.rs
ozwaldorf: Since pretty much all ip apis have very similar responses, its pretty easy to adjust the field names and continue.

crates/ursa/src/main.rs
b0xtch: This is pretty straightforward, we can get the total storage space/cpu/mem in rust and create a struct 

```
use sysinfo::{DiskExt, System, SystemExt};

#[derive(Debug, Serialize)]
pub struct Disk {
    name: String,
    total_space: u64,
    ...
}

pub struct SystemResources {
    total_ram: u64,
    disk: Disk,
    ....
}

impl SystemResources {
  fn system_resources(&self) -> Self {
     // use sysinfo to get all these resources
  }
}
```

crates/ursa-tracker/src/main.rs
ozwaldorf: This would be decent verification. Basically send a request to the provided address say for metrics and make sure it responds with 200?

Currently in the source there's no requirement for dns/tls to be setup. Should this be done separately or is that in scope for this pr?

crates/ursa-network/src/service.rs
```diff
@@ -11,40 +11,46 @@
...
-use libp2p::gossipsub::{MessageId, TopicHash};
-use libp2p::identify::Event as IdentifyEvent;
-use libp2p::multiaddr::Protocol;
-use libp2p::ping::Event as PingEvent;
-use libp2p::request_response::{RequestResponseEvent, RequestResponseMessage};
-use libp2p::swarm::{ConnectionHandler, IntoConnectionHandler, NetworkBehaviour};
-use libp2p::Multiaddr;
 use libp2p::{
-    gossipsub::IdentTopic as Topic,
+    autonat::{Event as AutonatEvent, NatStatus},
```
b0xtch, line 23: Was this automated or manually done?

ozwaldorf: https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#imports_granularity

I have it set locally but we could easily define a rustfmt.toml and enforce this formatting for everyone's `cargo fmt`

b0xtch: We don't currently enforce fmt, the default does a pretty good job and it's consistent across devs.

crates/ursa/src/main.rs
b0xtch: We can tackle this in another PR I am approving this for now

crates/ursa-tracker/src/main.rs
ozwaldorf: Will take care of this with another PR, specific to making changes to enforce DNS/TLS usage by node providers. Exact requirements can be up for discussion

crates/ursa-tracker/src/main.rs
ozwaldorf: Good catch! Sneaky unwrap. Though it's low priority as we are the ones running this tracker and it's setup through docker and proxies through nginx, which will always overwrite this header

b0xtch (approved):
Let's create issues from the comments made here 🙏

crates/ursa-metrics/Cargo.toml
```diff
@@ -2,13 +2,15 @@
...
 license = "MIT"
 repository = "https://github.com/Psychedelic/ursa"
 description = "Ursa metrics"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+prometheus = "0.13.2" # used by libp2p_bitswap's internal metrics
```
b0xtch, line 12: Could we use workspace dependencies here?

crates/ursa-metrics/src/server.rs
```diff
@@ -0,0 +1,54 @@
...
+    TextEncoder::new()
+        .encode(&BITSWAP_REGISTRY.gather(), &mut buffer)
+        .unwrap();
+    if !buffer.is_empty() {
+        metrics.push_str(&String::from_utf8(buffer).unwrap());
+    }
+
+    (StatusCode::OK, metrics)
+}
+
```
b0xtch, line 38: Could we consolidate the endpoints under one server?

infra/global-metrics/init-letsencrypt.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
b0xtch, line 1: This is a duplicate from ursa/ folder

b0xtch: Can pass create cars in place of urls for variability

infra/tf/configs/metrics.yml
```diff
@@ -0,0 +1,25 @@
+#cloud-config
```
b0xtch, line 1: Env variables can go in terraform, also are we installing docker inside a docker image?

crates/ursa-metrics/Cargo.toml
ozwaldorf: Sure, although it's only used by the single crate and not a common one.

crates/ursa-metrics/src/server.rs
ozwaldorf: hmm, I think this could be possible, but it'd require a refactor of both rpc and metrics I think. I had similar thoughts, and I think it'd be a good future improvement. I can track it in a new issue

infra/global-metrics/init-letsencrypt.sh
ozwaldorf: Its actually slightly different and doesn't run any prompts, because its used in the cloud config in terraform. I could maybe refactor for a single script that has a promptless option?

crates/ursa-metrics/src/server.rs
ozwaldorf: Actually was able to set this up pretty easily. I can push this now

ozwaldorf: https://github.com/fleek-network/ursa/pull/81/commits/ff6eb5ff97be3cde04aee8e3846939fd8af897dd

crates/ursa-metrics/Cargo.toml
ozwaldorf: https://github.com/fleek-network/ursa/pull/81/commits/e582e39196e44f3059b640f946f2eb6f638205b3

ozwaldorf:
redeploying the tracker compose with latest changes

infra/global-metrics/init-letsencrypt.sh
b0xtch: i just diffed the two from `/ursa` and this folder. They are exactly the same, aside from the domain name

b0xtch: the original script was intended to handle the `DOMAINS` var for any use case.

infra/ursa/init-letsencrypt.sh
```diff
@@ -1,24 +1,21 @@
...
   exit 1
 fi
 
-domains=(ursa.earth www.ursa.earth)
 rsa_key_size=4096
 data_path="./data/certbot"
 email="major@ursa.earth" # Adding a valid address is strongly recommended
 staging=1 # Set to 1 if you're testing your setup to avoid hitting request limits
 
-if [ -d "$data_path" ]; then
```
b0xtch, line 14: why are we deleting this section?

infra/ursa/init-letsencrypt.sh
```diff
@@ -27,33 +24,31 @@ if [ ! -e "$data_path/conf/options-ssl-nginx.conf" ] || [ ! -e "$data_path/conf/
   echo
 fi
 
-echo "### Creating dummy certificate for $domains ..."
```
b0xtch, line 30: why we capitalizing variable

infra/tf/configs/metrics.yml
```diff
@@ -0,0 +1,25 @@
...
+groups:
+  - ubuntu: [root,sys]
+
+runcmd:
+  - cd /root
+  - ufw allow 80
+  - ufw allow 443
+  - echo "IPINFO_TOKEN=${ipinfo_token}" >> /etc/environment
+  - source /etc/environment
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
```
b0xtch, line 11: most of the docker scripts here can be replaced with 

```
curl -fsSL https://get.docker.com -o get-docker.sh
sh get-docker.sh
```

infra/ursa/init-letsencrypt.sh
ozwaldorf: testing added it back with a option to skip via first arg

infra/ursa/init-letsencrypt.sh
ozwaldorf: its changed. But the variable now provided to the script is capitalized

infra/tf/configs/metrics.yml
ozwaldorf: https://github.com/fleek-network/ursa/pull/81/commits/e73d912b63a13d5be09c7146e7140440fdb25814

infra/tf/configs/metrics.yml
```diff
@@ -8,10 +8,8 @@ runcmd:
...
   - source /etc/environment
-  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
-  - add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
-  - apt-get update -y
-  - apt-get install -y docker-ce docker-ce-cli containerd.io
+  - curl -fsSL https://get.docker.com -o get-docker.sh
+  - sh get-docker.sh
   - systemctl start docker
   - systemctl enable docker
   - sudo curl -L "https://github.com/docker/compose/releases/download/1.27.4/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```
b0xtch, line 15: `curl -fsSL https://get.docker.com -o get-docker.sh` also handles installing compose.

crates/ursa-rpc-server/src/server.rs
```diff
@@ -32,19 +32,24 @@ where
...
+            if metrics.is_some() { " + metrics" } else { "" }
+        );
+
+        let rpc = Router::new()
             .merge(routes::network::init())
             .layer(Extension(self.rpc_server.clone()));
 
         let http = Router::new()
             .merge(http::routes::network::init::<S>())
+            .merge(metrics.unwrap_or_else(Router::new))
```
b0xtch, line 47:
```suggestion
            .merge(ursa_metrics::routes::init())
```

another PR, also food for through.

ozwaldorf: Goal was to have it easy to make optional in the future if wanted

infra/tf/configs/metrics.yml
ozwaldorf: https://github.com/fleek-network/ursa/pull/81/commits/1986f85315f019d0d13b4c9514ffb181ded1b502

b0xtch (approved):

kckeiks (approved):
Reviewed the rpc related code. Looks fine.

PR #95, opened 2022-12-07
Upgrade to libp2p v0.50 

kckeiks:


crates/ursa-network/Cargo.toml
```diff
@@ -10,6 +10,8 @@ description = "Ursa's libp2p implementation"
 [dependencies]
 anyhow = "1.0.56"
 async-fs = "1.6.0"
+async-std = { version = "1.11.0", features = ["attributes"] }
```
b0xtch, line 13: Do we need a async-std for any feature?

kckeiks: Sorry I forgot to remove this we moved to Tokio.

crates/ursa-network/src/behaviour.rs
```diff
@@ -301,6 +279,80 @@ impl<P: StoreParams> Behaviour<P> {
...
+
+    pub fn subscribe(&mut self, topic: &Topic) -> Result<bool, SubscriptionError> {
+        self.inner.subscribe(topic)
+    }
+
+    pub fn unsubscribe(&mut self, topic: &Topic) -> Result<bool, PublishError> {
+        self.inner.unsubscribe(topic)
+    }
 
     pub fn publish_ad(&mut self, public_address: Multiaddr) -> Result<()> {
```
theBeardA, line 357: all other events.push_back were replaced, did you miss this one or is it intentional?

kckeiks: Yes, I did. Will fix

crates/ursa-network/src/behaviour.rs
```diff
@@ -660,71 +696,76 @@ impl<P: StoreParams> Behaviour<P> {
...
+    fn poll(
         &mut self,
-        event: RequestResponseEvent<UrsaExchangeRequest, UrsaExchangeResponse>,
-    ) {
-        self.handle_request_response(event)
+        cx: &mut Context<'_>,
+        params: &mut impl PollParameters,
+    ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ConnectionHandler>> {
+        loop {
+            if let Some(event) = self.events.pop_front() {
```
kckeiks, line 735: So the service uses `publish_ad` to queue events so here we dequeue them and return. Let me know if you think this is fine. @theBeardA Thanks for catching that.

kckeiks: One concern with this is that events pushed by `publish_ad` could pass ahead of the queue and not let us poll the other inner behaviours. Hmm...

kckeiks: I don't see a function in `libp2p` API that could inject an event so it can be polled returned along with the other events. Another option is to make a "dummy": NetworkBehaviour struct that can be polled. Internally it will have the `events` queue. Thoughts?

kckeiks: I was thinking something like this.

```rust
struct Dummy {
  events: VecDeque
}

impl NetworkBehaviour for Dummy {
...
    fn poll() {
      if let Some(event) = self.events.pop_front() {
        return Poll::Ready(event)
      } 
      Poll::pending
    }
}
```

kckeiks: In this way, it will be polled "fairly"

theBeardA: Awesome. this makes sense. thanks

kckeiks: This might not be needed, after discussing with @b0xtch, these functions will get refactored and handling will be moved to the service.

b0xtch (approved):
Mint

PR #96, opened 2022-12-08
chore: upgrade behaviours

b0xtch:
- remove unnecessary behaviours
- use tokio transport which includes the new quic
- remove gossip and transport structs
- [x] todo: remove behaviour handlers from `behaviour.rs`
- [x] fix tests - #103


crates/ursa-network/src/tests/mod.rs
```diff
@@ -0,0 +1 @@
+mod service_tests;
```
ozwaldorf, line 1: no need for mod here if linking tests directly with path

b0xtch: Maybe you are referring to another case which I am not aware of. But here you def need to mod.

b0xtch: Can remove the mod.rs file

theBeardA (dismissed):

theBeardA (approved):

PR #99, opened 2022-12-08
fix: 🐛 bind docker container ports to host

heldrida:
## Why?

The docker-run would ultimately run the docker container without exposing the necessary ports to the host, thus this PR declares it.


ozwaldorf:
should also expose 4070 (metrics) and 6009 (p2p)

heldrida:
> 6009

included

PR #105, opened 2022-12-09
feat: adding test plans

kckeiks:


test-plans/ursa-network/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "ursa-network"
```
jsonsivar, line 2: Is the folder `/ursa-network/` necessary? Also maybe we call this package `ursa-network-tests`?

kckeiks: This branch/draft is just for experimenting with `testground`, the plan is for us (the devs on working on this) to write a set of tests. Since we need different binaries for each test plan, it makes sense to seperate them per crate but we should definitely attempt to consolidate before merging.

jsonsivar: Oh I see, I just saw the crates folder so that makes sense to have a folder here per binary

kckeiks:
@ozwaldorf @matthias-wright I pushed code for the ping test but I don't think we should be writing tests for things like ping, gossipsub and other libp2p protocols because we did not implement them. As the issue #114 indicates, we should test the following:
  - peers
  - routing
  - replication
  - traffic

Thoughts?

ozwaldorf:
> @ozwaldorf @matthias-wright I pushed code for the ping test but I don't think we should be writing tests for things like ping, gossipsub and other libp2p protocols because we did not implement them. As the issue #114 indicates, we should test the following:
>   - peers
>   - routing
>   - replication
>   - traffic
> 
> Thoughts?

I kinda agree here, although we need to test that these behaviors work inside the ursa context. That's why i was thinking we just have a general networking/libp2p testground crate that uses our internal behavior and tests the libp2p functionalities within. Simple regression to make sure no features are lost such as if a polling scheme or some event isnt propagated coreectly. For our custom behaviors and implementations we could create specific crates for each that do in depth tests

kckeiks:
> > @ozwaldorf @matthias-wright I pushed code for the ping test but I don't think we should be writing tests for things like ping, gossipsub and other libp2p protocols because we did not implement them. As the issue #114 indicates, we should test the following:
> > 
> > * peers
> > * routing
> > * replication
> > * traffic
> > 
> > Thoughts?
> 
> I kinda agree here, although we need to test that these behaviors work inside the ursa context. That's why i was thinking we just have a general networking/libp2p testground crate that uses our internal behavior and tests the libp2p functionalities within. Simple regression to make sure no features are lost such as if a polling scheme or some event isnt propagated coreectly. For our custom behaviors and implementations we could create specific crates for each that do in depth tests

Oh I understand now. I'll PM you.

kckeiks:
> > @ozwaldorf @matthias-wright I pushed code for the ping test but I don't think we should be writing tests for things like ping, gossipsub and other libp2p protocols because we did not implement them. As the issue #114 indicates, we should test the following:
> > 
> > * peers
> > * routing
> > * replication
> > * traffic
> > 
> > Thoughts?
> 
> I kinda agree here, although we need to test that these behaviors work inside the ursa context. That's why i was thinking we just have a general networking/libp2p testground crate that uses our internal behavior and tests the libp2p functionalities within. Simple regression to make sure no features are lost such as if a polling scheme or some event isnt propagated coreectly. For our custom behaviors and implementations we could create specific crates for each that do in depth tests

Testing our internal behaviour will also be easier once #96 is merged.

test-plans/ping/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+use db::{rocks::RocksDb, rocks_config::RocksDbConfig};
+use std::sync::Arc;
+
+#[async_std::main]
+async fn main() -> Result<()> {
+    let local_key = identity::Keypair::generate_ed25519();
+    let local_peer_id = PeerId::from(local_key.public());
+
+    //
+    let db = RocksDb::open("test_db", &RocksDbConfig::default())
```
theBeardA, line 21: just wanted to see if we can use memorydb instead rocksdb for tests?  we probably don't need to persist the data for test plans. I started using that in our unit test as well.
not sure if it directly applies but something like this:
https://github.com/fleek-network/ursa/blob/a38223782ed2b66b90930332d7cf8cbb2169e7d1/crates/ursa-network/src/service.rs#L880

kckeiks: Yes I agreed. Will use memorydb for test plans.

kckeiks:
I've been thinking about the high level algorithms for the tests. For example, I'm working on a test-plan that more or less goes like this:

1. First node that gets "initialized" in the network becomes the bootstrap node.
2. The second node joins the network and publishes some message.
3. We validate that all nodes subscribed to the topic receive the published message.

One issue with implementing this test is that we will need some way to access the state of the ursa service because virtually all of its state is private. Thoughts?
I have some ideas (not very good ones I'm afraid) so ping me if you'd like to discuss.

test-plans/ping/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+    let db = Arc::new(db);
+    let store = Arc::new(Store::new(Arc::clone(&db)));
+    //
+
+    let config = NetworkConfig::default();
+    // let index_provider = Provider::new(local_key.clone(), index_store, provider_config.clone());
+    let bitswap_store = BitswapStorage(store.clone());
+    // ---
+
+    let swarm = OrphanRuleWorkaround(Swarm::with_async_std_executor(
```
b0xtch, line 33: should we use tokio executor here?

kckeiks: Yes, will update.

kckeiks:
Moving this work to #130

PR #116, opened 2022-12-13
feat: parse home directories in config files, fix config defaults

### LINKED ISSUE #92: bug: improper path parsing for `~`
When using the config provided in the README, the paths specified `~/...` do not get parsed correctly and result in a folder created literally called `~`. Should parse this correctly as the HOME dir

ozwaldorf:
Putting a relative or home (`~`) path in the config should correctly resolve. 

- feat: use `resolve_path` crate on config flag, and all path fields,
- refactor: `ursa::config::load_config` actually loads and return the config, creating a default one if not found. 
- refactor: apply rpc_port flag in `CliOpts::to_config`
- refactor: Result type used with load_config (and related fns) to anyhow
- fix: uses of `env!("HOME")` in config defaults which only runs on compile time. Switch to `dirs::home_dir()` which works on runtime

fixes #92


crates/ursa/src/config.rs
```diff
@@ -1,26 +1,43 @@
...
+        let mut config: UrsaConfig = toml::from_str(&toml)?;
+        // parse relative and home directory paths
+        config.network_config.keystore_path =
+            config.network_config.keystore_path.resolve().to_path_buf();
+        config.network_config.database_path =
+            config.network_config.database_path.resolve().to_path_buf();
+        config.provider_config.database_path =
+            config.provider_config.database_path.resolve().to_path_buf();
+
+        Ok(config)
```
b0xtch, line 39: I would try to maintain `load_config` to be less verbose. We don't need to parse each individual field separately

ozwaldorf: moved the resolve calls to where the config values are instead of all at once

crates/ursa/src/config.rs
```diff
@@ -1,26 +1,33 @@
...
+        let toml = read_file_to_string(path)?;
+        toml::from_str(&toml).map_err(|e| anyhow!("Failed to parse config file: {}", e))
+    } else {
+        // Missing, create and return default config at path
         let ursa_config = UrsaConfig::default();
         let toml = toml::to_string(&ursa_config).unwrap();
         create_dir_all(path.parent().unwrap())?;
         let mut file = File::create(path)?;
-        return file.write_all(toml.as_bytes());
+        file.write_all(toml.as_bytes())?;
```
qti3e, line 28: use `.context()` to give a little context to the error messages for better diagnostics for users.

crates/ursa/src/ursa/mod.rs
```diff
@@ -55,38 +54,22 @@ pub struct CliOpts {
...
-                "Reading configuration from user provided config file {}",
-                config_file
-            );
-            path = PathBuf::from(&config_file);
+        let mut config = load_config(
+            &self
+                .config
+                .as_ref()
+                .map(|p| PathBuf::from(p).resolve().to_path_buf())
+                .unwrap_or_else(|| PathBuf::from(env!("HOME")).join(DEFAULT_CONFIG_PATH_STR)),
```
qti3e, line 62: `env!` macro is executed at compile time:

> Inspects an environment variable at compile time [[source](https://doc.rust-lang.org/std/macro.env.html)]

so this is not portable code and the generated binary will break on another machine.

crates/ursa/src/config.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/116/commits/7e776b95681a5a63c6776418a235945b9f27fb16

crates/ursa/src/ursa/mod.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/116/commits/7c838e19cbc3163ba3c4fce109b4f510b29135b6

crates/ursa/src/ursa/mod.rs
```diff
@@ -55,38 +55,22 @@ pub struct CliOpts {
...
-                "Reading configuration from user provided config file {}",
-                config_file
-            );
-            path = PathBuf::from(&config_file);
+        let mut config = load_config(
+            &self
+                .config
+                .as_ref()
+                .map(|p| PathBuf::from(p).resolve().to_path_buf())
+                .unwrap_or_else(|| PathBuf::from(var("HOME").unwrap_or_default()).join(DEFAULT_CONFIG_PATH_STR)),
```
qti3e, line 63: or you can just use [this crate](https://crates.io/crates/home) and be safe.

crates/ursa/src/config.rs
```diff
@@ -30,3 +37,17 @@ pub struct UrsaConfig {
...
     pub server_config: ServerConfig,
 }
+
+/// Read file as a `String`.
+pub fn read_file_to_string(path: &Path) -> Result<String> {
+    let mut file = match File::open(path) {
+        Ok(file) => file,
+        Err(error) => {
+            error!("Problem opening the file: {:?}", error);
+            process::exit(1);
```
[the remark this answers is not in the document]
ozwaldorf: fixed: https://github.com/fleek-network/ursa/pull/116/commits/060e1059fb0feef3dfb5bb19774c4fbe780b3a3c

crates/ursa/src/ursa/mod.rs
ozwaldorf: used `dirs::home_dir` which `resolve_path` uses internally. It has 34 million crate downloads so should be pretty safe

b0xtch (dismissed):

kckeiks (approved):

Issue #119, opened 2022-12-13
feat: gateway admin server


scott-dn:

scott-dn:
motivation:
- config management
- workers management
- cache management
...

scott-dn:
closed as #123

PR #107, opened 2022-12-12
feat: gateway query indexer

scott-dn:
POC: query global indexer

DEMO: 
```bash
$ curl https://localhost:80/bafybeigvgzoolc3drupxhlevdp2ugqcrbcsqfmcek2zxiw5wctk3xjpjwy -k | jq

{
  "MultihashResults": [
    {
      "Multihash": "EiDVNlzli2ONH3OslRv1Q0BRCKUCsERWs3RbthTVu6Xptg==",
      "ProviderResults": [
        {
          "ContextID": "AXESIO50esRu0SvbUfSGOzLTWfIff1S54seFI/PtyDuPNkzZ",
          "Metadata": "kBKjaFBpZWNlQ0lE2CpYKAABgeIDkiAg+IcjHXCOHHUf8wNiVtnvhYTPwL5Fqnnr7GyOLZp48R5sVmVyaWZpZWREZWFs9W1GYXN0UmV0cmlldmFs9Q==",
          "Provider": {
            "Addrs": [
              "/ip4/76.219.232.45/tcp/24001"
            ],
            "ID": "12D3KooWPNbkEgjdBNeaCGpsgCrPRETe4uBZf1ShFXStobdN18ys"
          }
        },
        ...
      ]
    }
  ]
}
```

PR #123, opened 2022-12-13
feat: gateway admin server

scott-dn:
Ref: #119 
Depend on: #107


crates/ursa-gateway/rustfmt.toml
```diff
@@ -0,0 +1,8 @@
+condense_wildcard_suffixes = true
```
b0xtch, line 1: is this just formatting for gateway. Should we maintain that across all crates? as we have one in the root folder

b0xtch (approved):

crates/ursa-gateway/rustfmt.toml
scott-dn: Def will do that in next PR, as it’s used nightly and might need more experiments.

b0xtch: i was mainly referring to how the formatting will differ from other crates. We should maintain consistency with other crates

scott-dn: Oh, that is mainly the import formats.
But yeah, I agree we should maintain all crates.

b0xtch (approved):

Issue #121, opened 2022-12-13
feat: gateway content cache


scott-dn:

scott-dn:
Motivation: 
- TLL Cache indexer result

scott-dn:
#132

scott-dn:
done in #139

scott-dn:
solved by #143

PR #132, opened 2022-12-15
feat: gateway content cache

scott-dn:
#121


b0xtch:
Just bringing comments here from the group, if gateways do cache content, then we should use cache control headers for invalidation. Also, another note regarding the extent of gateway nodes in the network. Cache content or relay content. The latter seems to cut down on the number of hops and would cost less in terms of bandwidth.

scott-dn:
>  should use cache control headers for invalidation
agreed 👍

PR #140, opened 2022-12-17
Add test plan 

kckeiks:
Replaces #130 

Adds test plan for `UrsaService` bootstrapping. See README for instructions on how to try it out.

The idea is that on future iterations when `UrsaService` is developed to support other functionality like for example data replication, we will be able to develop this test or add different tests as needed. We should be able to support adding tests for different versions released and in development.


ozwaldorf (approved):
LGTM. One thing could be adding test plans to our CI, but I don't think it's needed at this time. Once we have extensive tests we could have them run only on main branch merges and add ci badges to readme

b0xtch (approved):

PR #141, opened 2022-12-19
chore: workspace dependencies

theBeardA:
motivation: A few of the dependencies in multiple crates were using different versions. Recently, [the dependency inheritance](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace) was added to cargo. so this would be a good way of managing dependencies and update them faster as we move on.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,3 +19,65 @@ exclude = [
 lto = true
 opt-level = 'z'
 codegen-units = 1
+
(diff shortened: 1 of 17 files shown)


b0xtch (approved):
Mint

kckeiks:
This is great!

kckeiks (approved):

PR #142, opened 2022-12-19
chore: resolve clippy lints and fix CI to actually use clippy

ozwaldorf:
CI runner is doing clippy tests, but we have many failures across the project. This PR resolves all the clippy errors surfaced through `make clippy`.
I realized upon inspection that our CI runner has a make task labeled `clippy`, but instead configured to run `make check`, which isn't the same thing. This PR sets the CI to use `make ci` which runs check/fmt/clippy in one

## How
- refactor: resolve clippy linting issues across the project
  - `mod tests` now have `#[cfg(test)]` resolving unused fn lints
  - rpc had a duplicate nested module names, moved defs to mod.rs so now `rpc::rpc -> rpc`
  - numerous explicit function paths where an existing imports contain them
  - duplicate `ignore=todo` in protocol tests. Note: these should be real tests
  - unused variables now prefixed with `_`
  - use `anyhow::Result` instead of a std result with anyhow error attached
- refactor: removed a duplicate enum (index provider)
- set CI runner to use `make ci` which runs `check/fmt/clippy`
- remove `--locked` from `make [soft-]clippy` to fix ci not having a lock file
- add rustfmt to rust installer task
- fix codecov tasks getting skipped because of incorrect if check


ozwaldorf:
The CI runner is using rust `nightly` version, so clippy is being run with 1.66.0 nightly version of rust, whereas we defined the version to be 1.65.0 in our `rust-toolchain.toml` file. So locally clippy lints are passing but here they are failing

.github/workflows/rust.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         rust: [nightly]
         make:
           - name: clippy
-            task: "check"
+            task: "clippy"
```
b0xtch, line 20: Good catch I created a CI make command Abd forgot to edit it. Can we change to that

b0xtch (dismissed):

.github/workflows/rust.yml
ozwaldorf: https://github.com/fleek-network/ursa/pull/142/commits/6a22c1d55330fcca3f1bb0abe6ae6756195128d9

b0xtch (approved):

b0xtch (dismissed):

b0xtch (dismissed):

theBeardA (approved):
Nice cleanup!

Issue #139, opened 2022-12-16
feat(gateway): TLRFU cache content


scott-dn:
cache improvement to #121

scott-dn:
implement TLRFU for caching content

scott-dn:
#143

PR #143, opened 2022-12-19
feat: Gateway TLRFU cache

scott-dn:
see #139


crates/ursa-gateway/src/cache/tlrfu.rs
```diff
@@ -0,0 +1,469 @@
...
+                .get_mut(&data.freq)
+                .with_context(|| format!("[TLRFU]: Key: {k} not found at freq {}.", data.freq))?;
+            let key = lru.remove(&data.lru_k).await.with_context(|| {
+                format!(
+                    "[TLRFU]: Failed to remove LRU key: {} not found at freq {}.",
+                    data.lru_k, data.freq
+                )
+            })?;
+            lru.is_empty().then(|| self.freq.remove(&data.freq));
+            data.freq += 1;
```
kckeiks, line 61: We're assuming that this is very unlikely to overflow?

scott-dn: Yes, buckets are unlikely to overflow.

crates/ursa-gateway/src/cache/tlrfu.rs
```diff
@@ -0,0 +1,468 @@
...
+struct Data {
+    value: Vec<u8>,
+    freq: usize,
+    lru_k: usize,
+    _tll: u128,
+}
+
+pub struct Tlrfu {
+    store: HashMap<Arc<String>, Data>,
+    freq: BTreeMap<usize, Lru<usize, Arc<String>>>, // shrinkable
```
kckeiks, line 21: so the outermost map is for us to track the "frequency" with a counter and the inner map's (LRU) key is to track the least recently used?

kckeiks (approved):
LGTM. Nice tests.

crates/ursa-gateway/src/cache/tlrfu.rs
scott-dn: Exactly 💯

crates/ursa-gateway/src/admin/mod.rs
```diff
@@ -43,7 +51,9 @@ pub async fn start_server(config: Arc<RwLock<GatewayConfig>>) -> Result<()> {
 
     let app = Router::new()
         .route("/config", get(get_config_handler))
-        .layer(Extension(config));
+        .route("/purge-cache", post(purge_cache_handler))
```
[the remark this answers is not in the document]
b0xtch: this shouldn't happen manually though, at least through cache control

crates/ursa-gateway/src/config.rs
```diff
@@ -96,6 +103,10 @@ impl Default for GatewayConfig {
                  * mh_url: "https://cid.contact/multihash".into(),
                  */
             },
+            cache: CacheConfig {
+                max_size: 200_000_000,  // 200MB
```
b0xtch, line 107: maybe i missed but is this cache for content or index list?

scott-dn: It's content cache.

b0xtch: 200MB is enough for content cache?

b0xtch: https://github.com/fleek-network/ursa/blob/b1d6cd56911ac5174fc4c1126cf5cb27c5838408/infra/ursa/data/nginx/app.conf#L1

nginx on each node has 10GB

scott-dn: Sure, 10GB is not feasible for a development machine, it's just a default configuration for local development.
Let's come up with the configuration mode (dev, test, prod), sound good?

b0xtch: I mean on the gateway it could be even higher right, like even 3x that. Yeah lets develop for prod env. The simplest case is to make it variable and not const

scott-dn: Sure, that is a configuration variable, subject to change via `config.toml` or `cli argument`.

PR #144, opened 2022-12-19
docs: rename specs folder to doc

b0xtch:


ozwaldorf:
Should we rename `network-ing.md` to `networking.md` or just `network.md`

Issue #124, opened 2022-12-13
feat: gateway resolver


scott-dn:
resolve content from cache node

scott-dn:
solved by #158

PR #152, opened 2022-12-20
chore: consolidate rpc client and server crates into ursa-rpc-service

kckeiks:
Motivation: while working on #124 I tried importing rpc client to make calls to our rpc server. Then I realized I also needed to include rpc-server in my dependencies and import stuff from there. Since the client is in its own crate, this PR allows the client to include all necessary code to make calls to server without importing server code.


kckeiks:
I also thought about consolidating rpc-server and rpc-client so we'd just have one crate. Thoughts?

ozwaldorf:
> I also thought about consolidating rpc-server and rpc-client so we'd just have one crate. Thoughts?

Had the same exact thought there as well. Were you thinking of having something like:
```
ursa-rpc/
  - client/
  - server/
  - lib.rs
```

If it's a single crate it'd be nice to take advantage of feature flags as well for each respective use case

b0xtch (dismissed):

b0xtch (dismissed):
Mint

b0xtch (approved):

ozwaldorf (approved):
LGTM. Side note: the failed test is a false positive

PR #156, opened 2022-12-20
feat: add option to specify the port for cli rpc requests

matthias-wright:
This adds the option to specify the port for cli rpc requests. Sending rpc requests to different ports is useful for developing and debugging. This change is backwards compatible.
For example, to send a put request to port 4071:
```
ursa rpc put <file> --port 4071
```
Without the `--port` option, the port for the request will be taken from [here](https://github.com/fleek-network/ursa/blob/main/crates/ursa-rpc-server/src/config.rs), as before:
```
ursa rpc put <file>
```


kckeiks (approved):
LGTM

b0xtch:
We shouldn't couple ports with call handlers.

PR #158, opened 2022-12-20
feat: gateway query cache nodes

kckeiks:


b0xtch:
https://github.com/fleek-network/ursa/pull/156#issuecomment-1360277795

crates/ursa-gateway/src/server/route/api/v1/get.rs
```diff
@@ -88,7 +122,7 @@ pub async fn get_block_handler(
     // 1. maintain N workers keep track of indexing data
     // 2. cherry-pick closest node
     // 3. cache TTL
-    (StatusCode::OK, Json(json!(indexer_response)))
+    (StatusCode::OK, Json(json!({ "data": resp })))
```
scott-dn, line 125: can we improve it a bit? how about pipelining the response without loading it to the memory?

crates/ursa-gateway/src/server/route/api/v1/get.rs
```diff
@@ -76,6 +79,37 @@ pub async fn get_block_handler(
...
     debug!("received indexer response for {cid}:\n{indexer_response:?}");
 
+    let resp = match get_provider(&indexer_response) {
+        Some(addrs) => {
+            let mut addr_iter = addrs.iter();
+
+            loop {
+                match addr_iter.next() {
+                    Some(&addr) => {
+                        let server_config = ServerConfig::new(4069, addr.to_string());
```
scott-dn, line 89: we should read/parse `port` info from the `AddrInfo`.

crates/ursa-gateway/src/server/route/api/v1/get.rs
```diff
@@ -76,6 +79,37 @@ pub async fn get_block_handler(
...
+                        let params = NetworkGetParams { cid: cid.clone() };
+                        match ursa_rpc_client::functions::get_block(params, server_config).await {
+                            Ok(resp) => break resp,
+                            Err(_) => error!("request to RPC server failed"),
```
scott-dn, line 93: Possibility in the case of 404 that the cache node doesn't keep the content.

kckeiks:
> [#156 (comment)](https://github.com/fleek-network/ursa/pull/156#issuecomment-1360277795)

> ... can be taken up a level on client init not on every handler

Sorry, not sure what you mean here. Do you mean having a `Client` struct and with methods `get_block`, `get_file`, etc.?

b0xtch:
> > [#156 (comment)](https://github.com/fleek-network/ursa/pull/156#issuecomment-1360277795)
> 
> > ... can be taken up a level on client init not on every handler
> 
> Sorry, not sure what you mean here. Do you mean having a `Client` struct and with methods `get_block`, `get_file`, etc.?

an rpc port shouldn't be passed down from every handler because of the reasons in the comment above

crates/ursa-index-provider/src/provider.rs
```diff
@@ -160,7 +160,7 @@ where
 
     fn create_announce_message(&mut self, peer_id: PeerId, mut domain: String) -> Result<Vec<u8>> {
         if domain.is_empty() {
-            domain = "/ip4/127.0.0.1/tcp/8070".to_string();
+            domain = "/ip4/127.0.0.1/tcp/4069".to_string();
```
b0xtch, line 163: @theBeardA

crates/ursa-gateway/src/server/route/api/v1/get.rs
kckeiks: Talked to Scott about this and we will optimize as part of a separate pr.

crates/ursa-gateway/src/server/route/api/v1/get.rs
kckeiks: Passing 404 response now

crates/ursa-gateway/src/server/route/api/v1/get.rs
kckeiks: @scott-dn After #170 I will rebase and update this.

crates/ursa-index-provider/src/provider.rs
theBeardA: yea, this should be default index provider port and not server port

kckeiks: You mean this should not be 4069 e.g. the RPC server port? 

hmm when the gateway queries the cache nodes, it needs to do it at the endpoint with this port. @scott-dn comments?

scott-dn: @kckeiks should be the advertisement address:
> https://github.com/ipni/storetheindex/blob/main/api/v0/ingest/schema/schema.ipldsch#L45-L46

kckeiks: @theBeardA you're right. This is the wrong place to update the address of where to retrieve the content. Reverted.

crates/ursa-gateway/src/server/route/api/v1/get.rs
```diff
@@ -96,7 +157,13 @@ pub async fn get_block_handler(
     // 1. maintain N workers keep track of indexing data
     // 2. cherry-pick closest node
     // 3. cache TTL
-    (StatusCode::OK, Json(json!(indexer_response)))
+    (
```
b0xtch, line 160: I am sure y’all are already thinking about it but we should be steaming back data.

b0xtch (approved):

scott-dn (approved):

crates/ursa-gateway/src/indexer/mod.rs
```diff
@@ -1 +1,56 @@
...
+
+use crate::indexer::model::IndexerResponse;
+use libp2p::multiaddr::Protocol;
+use rand::Rng;
+use std::net::SocketAddrV4;
+use tracing::{error, warn};
+
+// Randomly chooses a provider and returns its addresses.
+pub fn get_provider(indexer_response: IndexerResponse) -> Option<Vec<SocketAddrV4>> {
+    // We expect that MultihashResults will have at most 1 element.
```
b0xtch, line 11: @theBeardA iirc you can have multiple

kckeiks: @scott-dn

kckeiks: I think Scott said this would only happen if we batched content

scott-dn: @b0xtch`findCid` only returns one, `findBatch` can have multiple.
> https://github.com/ipni/storetheindex/blob/main/server/finder/http/handler.go#L57-L83

theBeardA: let me clarify, multihash element can have at most one element, that's correct.

crates/ursa-gateway/src/indexer/mod.rs
```diff
@@ -1 +1,56 @@
...
+    let providers = match indexer_response.multihash_results.get(0) {
+        Some(result) => &result.provider_results,
+        None => {
+            error!("indexer result did not contain a multi-hash result");
+            return None;
+        }
+    };
+
+    if !providers.is_empty() {
```
theBeardA, line 20: should we also filter by metadata "FleekNetwork" ?
https://github.com/fleek-network/ursa/blob/9d8d626aa414c8af03233f61749eb94549df1a4c/crates/ursa-index-provider/src/advertisement.rs#L54

crates/ursa-gateway/src/indexer/mod.rs
scott-dn: all goods @theBeardA 👍

crates/ursa-gateway/src/indexer/mod.rs
scott-dn: @theBeardA Yeah, true, good catch.

@kckeiks Maybe let me handle the filtering in another PR.

kckeiks: Will handle in different PR

theBeardA (approved):

crates/ursa-gateway/src/server/route/api/v1/get.rs
```diff
@@ -84,6 +87,92 @@ pub async fn get_block_handler(
...
+                    Some(addr) => {
+                        let req = Request::builder()
+                            .method("POST")
+                            .uri(format!("http://{}:{}/rpc/v0", addr.ip(), addr.port()))
+                            .header("Content-Type", "application/json")
+                            .body(Body::from(
+                                json!({
+                                    "jsonrpc": "2.0",
+                                    "id": "id",
+                                    "method": "ursa_get_cid",
```
theBeardA, line 105: this method only gets a single block. is this what this handler is trying to do, to just get a single block?
if it is meant to cache/get the full content, it should call the stream(http) api

kckeiks: @theBeardA I believe this is only to get a block per the handler function name and the todo comments but I could be wrong. @scott-dn is it?

scott-dn: @kckeiks as the gateway, we want full content, but that needed a streaming handler which we need atm (#198).
For this PR, let's assume we only get a single block.
hope it makes sense to you @theBeardA .

theBeardA: sure, we wanted to remove that method, since it's kinda useless, once we have more rpc methods like getting blockchain info. just wanted to make note of that

kckeiks: Good to know. Then it sounds like we def need a way to pipe the responses then.

scott-dn (approved):

PR #162, opened 2022-12-21
feat: periodic kad bootstraps

### LINKED ISSUE #73: chore: automatic kademlia walks
Nodes should perform a kademlia query every few minutes to discover new peers

ozwaldorf:
## Why

Nodes should periodically run a bootstrap query to actively discover new/closer peers. The interval should reset after the previous query completes. If more than a max num of peers is connected, don't attempt to discover more

## What

- First bootstrap has an initial delay of 5s
- New network config field `bootstrap_interval` (default 5 mins)
- Run a bootstrap if connected peers < `mesh_n` and an interval has passed
- set next bootstrap after previous query completes or an error
- refactor kad handler and improve logs

## Demo

image

## Where

resolves #73


crates/ursa-network/src/discovery.rs
```diff
@@ -91,16 +107,9 @@ impl DiscoveryBehaviour {
...
-        if let Err(error) = kademlia.bootstrap() {
-            warn!("Failed to bootstrap with Kademlia: {}", error);
-        }
-
-        let mdns = if config.mdns {
-            Some(Mdns::new(Default::default()).expect("mDNS start"))
-        } else {
-            None
-        };
+        let mdns = config
```
b0xtch, line 110: Revert

ozwaldorf: Sure thing, although was making it consistent with the usage of `.then_some()` a few lines up for the next_bootstrap

b0xtch: Can you make that one consistent with this?

ozwaldorf: https://github.com/fleek-network/ursa/pull/162/commits/1fca91c9621f84a47e18395fb8b17b65bc97b4f2

kckeiks:
This feature is a good candidate for validation in the bootstrap test plan. It would be useful to update it.

ozwaldorf:
to be refactored after #177 is complete

b0xtch:
Do we still need this, or is this changing to jar random walks otherwise we can close

Issue #151, opened 2022-12-20
feat: DNS + SSL requirements, verify registration, docker setup


ozwaldorf:
Nodes should be required to have a valid SSL certificate setup on a DNS address and listen over https. Once ssl is enforced we can do a simple verification on the tracker by making a valid request to the registering node's `/ping` endpoint  

## TODO:

- [ ] finalize config entry for dns (`node.domain` - rpc/http/metrics/index provider)
- [ ] SSL cert/key config entries OR proxy config entries if not direct https
- [ ] nodes include dns address for tracker registration
- [ ] tracker should assume the supplied address and 80/443 are available for rpc/http
- [ ] tracker makes https request to `/ping` on registering node to verify setup

## Open Questions:

- Will dns+ssl be a hard requirement to run the binary?
> no, because those who run the node behind nginx proxy with ssl and dont need the binary to handle ssl directly
- should nodes register tracker with a full url+port for http to support proxied nodes?
> tracker should automatically assume node.address:80/443 is available

b0xtch:
https://github.com/fleek-network/ursa/issues/53#issue-1337497746

related

b0xtch:
This can be split up into different issues, especially the ones that aren't currently actionable

PR #163, opened 2022-12-21
chore: update ursa docker compose to latest grafana and expose all ports

ozwaldorf:
## What

fixes for `infra/ursa/docker-compose.yml`:

- update grafana version
- expose proper ports and update comments

### DIFF
--- a/infra/ursa/docker-compose.yml
+++ b/infra/ursa/docker-compose.yml
@@ -76,8 +76,6 @@ services:
     restart: on-failure
     ports:
       - "6009:6009"
-      - "4069:4069"
-      - "8070:8070"
     expose:
       # 4069 TCP, used for HTTP RPC, REST, and metrics
       - "4069"
--- a/infra/ursa/docker-compose.yml
+++ b/infra/ursa/docker-compose.yml
@@ -51,7 +51,7 @@ services:
       - ursa
 
   grafana:
-    image: grafana/grafana:6.7.2
+    image: grafana/grafana
     volumes:
       - ./data/grafana:/var/lib/grafana
       - ./data/grafana/provisioning:/etc/grafana/provisioning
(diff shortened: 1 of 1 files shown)


b0xtch (dismissed):

infra/ursa/docker-compose.yml
```diff
@@ -75,13 +75,12 @@ services:
       dockerfile: Dockerfile
     restart: on-failure
     ports:
-      # 6009 TCP and UDP, used by the P2P protocol running the network
       - "6009:6009"
+      - "4069:4069"
```
b0xtch, line 79: 4069 doesn’t need to be exposed, as we use nginx in front

ozwaldorf: Hmm this actually opens an interesting problem: nodes currently register with the tracker with their raw internal port, there's no consideration for the proxy or https atm. I think this works temp until #151 is completed, which would make the necessary changes here. Also, the prometheus container is scraping ursa:4069, so it needs to be exposed for that internally.

b0xtch: We can draft then until that's ready

b0xtch: Nodes should register their domain ie tls

b0xtch: Using ports exposes on the Host you can scrape with no problem

ozwaldorf: > Using ports exposes on the Host you can scrape with no problem

Ah this is the missing link, perfect. So I can just remove the ports and leave expose

b0xtch (approved):

infra/ursa/docker-compose.yml
ozwaldorf: https://github.com/fleek-network/ursa/pull/163/commits/283935721376edbf60cbb18478de76f7325c19b8

PR #170, opened 2022-12-21
chore: refactor rpc client api

kckeiks:
Based on feedback in #156 and #158.


crates/ursa-rpc-service/src/client/mod.rs
```diff
@@ -1,4 +1,121 @@
...
+pub type Result<T> = anyhow::Result<T, Error>;
+
+pub struct Client {
+    server_addr: Ipv4Addr,
+    server_port: u16,
+}
+
+impl Default for Client {
+    fn default() -> Self {
+        Client::new(Ipv4Addr::from([0, 0, 0, 0]), 4069)
```
ozwaldorf, line 22: can we use a `const DEFAULT_PORT` to have this in once place consistent with the ServerConfig default

crates/ursa-rpc-service/src/client/mod.rs
```diff
@@ -1,4 +1,121 @@
...
+                    code: error.code,
+                    message: error.message,
+                }),
+            }
+        } else {
+            error!("[RPCClient] - There was an while serializing the rpc request");
+            Err(Error::Full {
+                data: None,
+                code: 200,
+                message: "There was an while serializing the rpc request".to_string(),
```
ozwaldorf, line 117:
```suggestion
            error!("[RPCClient] - There was an error while serializing the rpc request");
            Err(Error::Full {
                data: None,
                code: 200,
                message: "There was an error while serializing the rpc request".to_string(),
```

ozwaldorf (approved):
lgtm

crates/ursa/src/ursa/rpc_commands.rs
```diff
@@ -34,7 +32,11 @@ impl RpcCommands {
                 let params = NetworkPutFileParams {
                     path: path.to_string(),
                 };
-                match put_file(params, *port).await {
+                let mut client = Client::default();
```
b0xtch, line 35: Shouldn’t we initiate client per rpc service and not per command. Also noticed we are still passing rpc port in reach command

kckeiks: Since the feedback on #158 was on parameters of rpc client functions #158, command refactoring was not addressed here :(. I can open an issue for removing the ports though.

kckeiks: https://github.com/fleek-network/ursa/issues/179

b0xtch: This issue was introduced in #156 where the initial feedback was given

PR #175, opened 2022-12-22
improve: test cases

theBeardA:


.github/workflows/rust.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           - name: check-fmt-clippy
             task: "ci"
           - name: test
-            task: "test-network"
+            task: "test"
```
ozwaldorf, line 22: can remove the make entry `test-network` since we're not using it anymore

ozwaldorf:
service test put and get is failing:
image

Cargo.toml
```diff
@@ -26,7 +26,7 @@ async-fs = "1.6.0"
 async-trait = "0.1.60"
 axum = { version = "0.6.1", features = ["multipart", "headers"] }
 axum-server = { version = "0.4.4", features = ["tls-rustls"] }
-base64 = "0.20.0"
```
b0xtch, line 29: We going down down down?

theBeardA: yes, I upgraded that in workspace dependency branch and it broke few things. I will open another issue to upgrade this.

b0xtch:
https://github.com/fleek-network/ursa/blob/b92fc66eddf292adb874b39c3b546c38e3dd5637/.github/workflows/rust.yml#L22

Could we also update the test here to all now 🙏

b0xtch (approved):

PR #177, opened 2022-12-22
refactor: kad, mdns, and network tests

### LINKED ISSUE #154: tests(network): Fix false positives in kad, bitswap tests
Kad is a full false positive because it has the mdns behavior enabled. We should run a local bootstrap A and 2 nodes B and C. Connect B -> A, and C -> A, and let kad discovery connect B -> C.

With bitswap_get/sync tests, the nodes sometimes connect to bootstrap nodes faster than they connect to eachother. There is no check the right peer was connected, and so the test fails because the content id cannot be found as the only node connected is not the test node with the content. This is inconsistent though so sometimes test passes, sometimes it fails.

Side note:
Gossip tests are using the testnet bootstrap nodes. Do we want the tests connecting to our actual network or just local?

ozwaldorf:
## Why

Based on discussion with @b0xtch, the level of seperation/features provided by `discovery.rs` are no longer needed, so we should merge kad and mdns behaviours from there into `behaviour.rs`. This PR also refactors the network tests, of which there were false positives and other minor issues.

## What

- merge discovery directly into behaviour and service
- Slimmed down feature-set from discovery
- new behaviour::Behaviour::add_address
- only perform bootstrap if we're not a bootstrapper and have some nodes
- re-add peer connected and disconnected logs
- all test loops now send the events to the service handler properly to avoid unhandled events
- actual kademlia test with a local test bootstrap
- mdns test no longer uses bootstraps
- prefer `Behaviour(Identify)` over `ConnectionEstablished` for verifying test node connection for most tests
- replaced SimpleLogger with tracing_subscriber
- new `tag` helper fn to add a node nametag to logs for each node in a test

## Notes

> network bitswap tests are still unstable. They will need to be further debugged separately, but the changes made seem to improve them slightly

closes #154


crates/ursa-network/src/behaviour.rs
```diff
@@ -159,10 +164,40 @@ impl<P: StoreParams> Behaviour<P> {
                 if relay_client.is_none() {
                     panic!("relay client not instantiated");
                 }
-                dcutr::behaviour::Behaviour::new()
+                Dcutr::new()
             })
             .into();
 
+        let mdns = config
```
b0xtch, line 171: Revert

crates/ursa-network/src/tests/service_tests.rs
```diff
@@ -21,26 +20,35 @@ mod tests {
         Multiaddr, PeerId,
     };
     use libp2p_bitswap::BitswapStore;
-    use simple_logger::SimpleLogger;
+    use std::fmt::Display;
```
b0xtch, line 23: Revert all

crates/ursa-network/src/behaviour.rs
ozwaldorf: all the other variables in this function are using `then(||` syntax, what do?

b0xtch: Close the pr

ozwaldorf:
?

I assume you have issues with the tag and logging in the tests refactoring. This was immensely helpful for debugging the kad and mdns tests when looking into them, so I included them in the pr. But I can revert. 

In terms of the event handling and loops; this is required since we dont start the service loop that passes events into the handler automatically. Also the nodes other than the one we want should still run so our swarm handler gets all events correctly. The tests that did pass without handling events worked but only by the default behaviours, none of our own logic. this ensures our service code is accounted for while node runs

Issue #176, opened 2022-12-22
Bug: Docker setups for running nodes do not save data between runs


ozwaldorf:
`~/.ursa/` should be linked as a volume to the docker instance for make scripts, and in the getting started documentation. Users that run the node this way lose their keys and other cache data between instances of the node

b0xtch:
not a bug, we have set it up to bare bones. Ideally, you should use the docker-compose setup for persistence

PR #178, opened 2022-12-22
fix: make docker-run persistence

ozwaldorf:
## Why?

#176 

## What?

- add `-v ~/.ursa:/root/.ursa` to make script to launch docker with persistence
- refactor config default to use `~` and parse later vs `home_dir()` directly. This eases config compatibility when running the node directly vs in docker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ docker-build:
 docker-run:
 	docker run --name ursa-cli \
 		-p 4069:4069 -p 6009:6009 -p 8070:8070 \
-		-v ~/.ursa:~/.ursa -it ursa
+		-v ~/.ursa:/root/.ursa -it ursa
 
 compose-up:
 	docker-compose -f infra/ursa/docker-compose.yml up
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,9 @@ docker-build:
 	docker build -t ursa -f ./Dockerfile .
 
 docker-run:
-	docker run -p 4069:4069 -p 4070:4070 -p 6009:6009 -p 8070:8070 --name ursa-cli -it ursa
+	docker run --name ursa-cli \
+		-p 4069:4069 -p 6009:6009 -p 8070:8070 \
+		-v ~/.ursa:~/.ursa -it ursa
 
 compose-up:
 	docker-compose -f infra/ursa/docker-compose.yml up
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -19,7 +19,9 @@ docker-build:
 	docker build -t ursa -f ./Dockerfile .
 
 docker-run:
-	docker run -p 4069:4069 -p 4070:4070 -p 6009:6009 -p 8070:8070 --name ursa-cli -it ursa
+	docker run --name ursa-cli \
```
b0xtch, line 22: we should add that to the docker file instead

ozwaldorf: How so? 
- VOLUME in dockerfile only saves a volume generated on build, so it's not the same as passing flag
- we only have EXPOSE in dockerfile, if used ports should still be open to host for custom nginx setups etc right?

b0xtch: Why would you want it updated in real time? What if a user doesn’t want that? Command should not be explicit to serve all needs. I would also remove the ports there

ozwaldorf: Why save the ursa dir?
- Pem file for the nodes - this be a bigger deal with consensus
- cached content is supposed to be held for 30 days - again bigger deal once nodes have the hard requirement

If a user restarts the node without persistence, they get a completely new environment from scratch

b0xtch: We don't have consensus

b0xtch: Also I suggest reading adding volumes to docker file https://docs.docker.com/storage/volumes/

ozwaldorf: okay, what about the existing docker-compose? This is just making the `make docker-run` exactly the same as `make compose-up`

b0xtch: Docker compose is the use case, the container is the tool. Not intended way to run a prod ready node. Still for persistence we don't need the explicit command there read above linked

ozwaldorf: Our documentation needs to be updated then if that's the case; most of the "getting started with docker" blog is about running the dockerfile or our published image directly, and only a mention at the end about the compose setup

b0xtch: nonissue, should we close?

ozwaldorf: I still think theres still a problem with not keeping it persistent. Because some users are updating their nodes hourly and running under new peer ids each time. For example, on the tracker there is one node that's registered with 33 different peer ids today

b0xtch: Yeah let's update blogs stating to use docker compose or the one you mentioned. 

On the registration, peer ids can change the tls certificate in the future is far the only mapping that can't change

PR #180, opened 2022-12-22
refactor: merge discovery directly into behaviour and service

ozwaldorf:
- minimal logic used from discovery implemented in behaviour and 
- unified `add_address` logic throughout behavior.rs 
- only perform bootstrap if we're not a bootstrapper and have some nodes
- use verbose paths for QueryId to increase readability as we now have both `kad::QueryId` and `bitswap::QueryId` in behaviour.rs


ozwaldorf:
Note: I didn't include fixes for the kad test with this, so it's still a false positive as mentioned to refactor tests with another PR separately

crates/ursa-network/src/behaviour.rs
```diff
@@ -159,10 +164,40 @@ impl<P: StoreParams> Behaviour<P> {
                 if relay_client.is_none() {
                     panic!("relay client not instantiated");
                 }
-                dcutr::behaviour::Behaviour::new()
+                Dcutr::new()
             })
             .into();
 
+        let mdns = config
```
b0xtch, line 171: Revert

crates/ursa-network/src/service.rs
```diff
@@ -196,6 +198,10 @@ pub struct UrsaService<S> {
     _pending_requests: HashMap<RequestId, ResponseChannel<UrsaExchangeResponse>>,
     /// Pending responses.
     pending_responses: HashMap<RequestId, oneshot::Sender<Result<UrsaExchangeResponse>>>,
+    /// Connected peers.
+    peers: HashSet<PeerId>,
```
b0xtch, line 202:
```rust
    /// Information about connected peers.
    peer_info: HashMap<PeerId, Vec<Multiaddr>>,
```

if are not using peer_info where are we strong the mapping of the `peer_id` and their information

ozwaldorf: It's not needed anymore, all the behaviours track this themselves internally and can be accessed at any point with `behaviour.get_peer_addresses()`. This includes addresses from every behaviour internally and you can verify by running a `cargo expand` to view the macro provided implementation

crates/ursa-network/src/service.rs
```diff
@@ -491,19 +492,50 @@ where
...
-            }
-            DiscoveryEvent::Disconnected(peer_id) => {
-                trace!(
-                    "[DiscoveryEvent::PeerDisconnected] - Peer disconnected {:?}",
-                    peer_id
-                );
-                self.emit_event(NetworkEvent::PeerDisconnected(peer_id));
+    pub fn handle_kad(&mut self, event: KademliaEvent) -> Result<()> {
+        match event {
+            KademliaEvent::OutboundQueryProgressed { id, result, .. } => match result {
```
b0xtch, line 497: Can we handle these kad events

```
GetProviders
GetClosestPeers
```
This is where you can use the peer info

ozwaldorf: > GetProviders

This wouldn't ever get generated as we're not using kad's providers or querying for them

> GetClosestPeers

In the current state this event is never generated (it's generated from `behaviour.kad.get_closest_peers()`), but I plan on using + handling it for random kad walks

crates/ursa-network/src/service.rs
b0xtch: Yeah that make sense given i suggested the move away form having a custom handler still how do we handle custom peer info

ozwaldorf: > custom peer info

I didn't see any actual usages of the peer_info struct other than insertions, so I don't think we need to worry for now. What are you anticipating we'd need to store?

b0xtch: When we handle custom gossip sub logic with scoring, content routing, and more. You can leave it out, for now, I guess.

crates/ursa-network/src/service.rs
b0xtch: That make sense

b0xtch (approved):

crates/ursa-network/src/behaviour.rs
ozwaldorf: Done. But code is now inconsistent

image

b0xtch (approved):
Mint

PR #184, opened 2022-12-23
tests: kad and mdns

### LINKED ISSUE #154: tests(network): Fix false positives in kad, bitswap tests

ozwaldorf:
depends on #180

---

## Why

Kad and mdns tests are not truely testing what they say, and now also endlessly looping.

We must either run `UrsaService::start()` or if not, manually run `UrsaService::handle_swarm_event` to actually have our service logic executed. Since discovery behaviours are no longer executed through poll() and instead just surfaced as events to the service, the mdns discovery isn't handled and causes a loop since the dial never happens. Previously kad was a false positive due to mdns, but since mdns isn't correctly run, it also loops.

## What

- refactor mdns test to not need bootstrap nodes anymore and correctly handle swarm events.
  - eliminates the chance of kad connecting the nodes when we're testing for mdns
- refactor discovery test to an actual kad test:
  - Bootstrap, Node 1, and Node 2:
    1. Node 1 connects to Bootstrap, waits for identify
    2. Node 2 connects to Bootstrap
    3. Wait for Node 2 to connect to Node 1

## Related Issues

closes #154


crates/ursa-network/src/tests/service_tests.rs
```diff
@@ -191,42 +189,71 @@ mod tests {
...
+            }
+        }
+
+        let (mut node_2, ..) =
+            network_init(&mut NetworkConfig::default(), Some(bootstrap_addr), None).await?;
+
+        // wait for node 2 to connect with node 1 through kad peer discovery
+        loop {
+            select! {
+                event = node_1.swarm.select_next_some() => node_1.handle_swarm_event(event)?,
```
b0xtch, line 246: Let’s in-line the event handler here for visibility

ozwaldorf: Not sure what you mean?

b0xtch: Like similar to the other tests where we are explicit about the event being handled

ozwaldorf: We're not looking for events in node 1, only node 2, so we just pass to the service handler to let the node do it's thing in the background. Otherwise, the mdns events aren't handled and the nodes don't dial eachother

b0xtch: Same comment still applies, instead of calling the handle swarm events, we should inline for visibility and consistency with other tests

ozwaldorf: other tests just discard the event, and none of the service.rs logic is actually used. With mdns if you just disable bootstrap nodes and don't make any other modifications, it'd loop endlessly because the mdns handler never gets the event so the dial never happens etc. Since the refactor in #96 it's been doing this because the internal behavior poll() doesn't execute the code. Also some tests still work because start() is used, which loops and passes event to swarm_event_handler

b0xtch: I am not sure I am following. When you call start the event loop gets started. Which is what should happen. My focus is on inclining the event handler and not using the handle_event function

ozwaldorf: Yes, since we're selecting events before calling start, the actual event loop that calls handle_event is not present. And since kad and mdns logic is now included in the service.rs, it doesn't work unless you explicitly allow it to

b0xtch: I am not sure you have a full picture of how the event loop works. I'll suggest a change once I am back on my laptop

ozwaldorf: Sure thing, but just look at tests on main now, had to cancel because endless loop as events aren't being handled. I manually debugged and verified this earlier today

b0xtch: For discovery?

b0xtch: Event handling is best to be handled inlined for visibility

ozwaldorf: yeah both kad and mdns are endless looping

b0xtch: Mdns takes time to be handled, discovery can offer false positives.

ozwaldorf: Maybe a util method that wraps select_next_some() and passes to handler, and then returns the event, instead of just returning next event would be a good improvement. Then start could use it, and we could use it in tests instead of select_next_some to keep the code less verbose

b0xtch: We are not looking to rewrite tests atm, but could be a good future improvement

ozwaldorf: Does this resolve what you wanted? https://github.com/fleek-network/ursa/pull/184/commits/77d4891d1759815804c13da8f0933d1979913018

ozwaldorf: https://github.com/fleek-network/ursa/pull/184/commits/c84f5e1a9ec7ddd4ec00d98fd5571b2fbeb942f6

crates/ursa-network/src/tests/service_tests.rs
```diff
@@ -61,21 +61,20 @@ mod tests {
         }
     }
 
-    async fn _run_bootstrap(
+    async fn run_bootstrap(
```
b0xtch, line 64: we don't need to start two nodes for this test; one is fine to test that the network `inits`

crates/ursa-network/src/tests/service_tests.rs
```diff
@@ -61,21 +61,20 @@ mod tests {
...
-        let (bootstrap, ..) = network_init(config, Some(addr.clone()), Some(keypair)).await?;
-        Ok((bootstrap, addr))
+        config.swarm_addrs = vec!["/ip4/127.0.0.1/tcp/0".parse().unwrap()];
+        config.bootstrapper = true;
+        config.bootstrap_nodes = vec![];
+        let (bootstrap, addr, peer_id, ..) = network_init(config, None, Some(keypair)).await?;
+        Ok((bootstrap, addr, peer_id))
     }
 
     async fn network_init(
```
b0xtch, line 75: we don't need to start two nodes for this test; one is fine to test that the network inits

crates/ursa-network/src/tests/service_tests.rs
```diff
@@ -182,50 +180,72 @@ mod tests {
...
-            mdns: true,
-            ..Default::default()
-        };
 
-        let (mut node_1, _, peer_id_1, ..) = network_init(&mut config, None, None).await?;
-        let (mut node_2, ..) = network_init(&mut config, None, None).await?;
+        let (bootstrap, bootstrap_addr, bootstrap_id) =
+            run_bootstrap(&mut NetworkConfig::default()).await?;
 
+        // no need to loop over bootstraps events, we can just start it up
```
b0xtch, line 211: verbosseeeeeeeeee

b0xtch: 🙃

crates/ursa-network/src/tests/service_tests.rs
```diff
@@ -182,50 +180,72 @@ mod tests {
...
+            )) = node_1.swarm.select_next_some().await
+            {
+                info!("[SwarmEvent::Identify::Sent]: {peer_id:?}, {bootstrap_id:?}");
+                if peer_id == bootstrap_id {
+                    break;
+                }
+            }
+        }
+
+        // let node 1 run in the background
```
b0xtch, line 234: verbose

b0xtch:
Looks good

b0xtch (dismissed):

b0xtch (approved):

PR #185, opened 2022-12-23
fix: advertise server address

kckeiks:
Blocks #158 

**Motivation**
The address of the server from which advertised content is retrievable wasn't being advertised.

### DIFF
--- a/crates/ursa-index-provider/src/engine.rs
+++ b/crates/ursa-index-provider/src/engine.rs
@@ -90,6 +90,8 @@ pub struct ProviderEngine<S> {
     command_receiver: Receiver<ProviderCommand>,
     /// network command sender for communication with libp2p node
     network_command_sender: Sender<NetworkCommand>,
+    /// Server from which advertised content is retrievable.
+    server_address: Multiaddr,
 }
 
 impl<S> ProviderEngine<S>
@@ -102,6 +104,7 @@ where
         provider_store: Arc<UrsaStore<S>>,
         config: ProviderConfig,
         network_command_sender: Sender<NetworkCommand>,
+        server_address: Multiaddr,
     ) -> Self {
         let (command_sender, command_receiver) = unbounded_channel();
         ProviderEngine {
@@ -111,6 +114,7 @@ where
             network_command_sender,
             provider: Provider::new(keypair, provider_store),
             store,
+            server_address,
         }
     }
     pub fn command_sender(&self) -> Sender<ProviderCommand> {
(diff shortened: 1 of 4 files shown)


crates/ursa-index-provider/src/engine.rs
```diff
@@ -208,7 +212,7 @@ where
         let peer_id = PeerId::from(self.provider.keypair().public());
 
         let listener_addresses = listener_addresses_receiver.await?;
```
kckeiks, line 214: Do we need to advertised these other addresses that we're getting for the swarm? @theBeardA

crates/ursa/src/main.rs
```diff
@@ -96,13 +97,20 @@ async fn main() -> Result<()> {
...
+                .expect("Server to have a valid address");
+
                 let index_store = Arc::new(UrsaStore::new(Arc::clone(&Arc::new(provider_db))));
                 let index_provider_engine = ProviderEngine::new(
                     keypair,
                     Arc::clone(&store),
                     index_store,
                     provider_config,
                     service.command_sender(),
+                    server_address,
```
kckeiks, line 113: Any other suggestions on how to pass the address? 🤔

crates/ursa-index-provider/src/engine.rs
scott-dn: I don't think so; we only need to advertise where the content is retrievable, which is the rpc server.

theBeardA: The indexers filters out all the local addresses. So there is no problem adding them to the list. When ran locally, most of the time public address is not available, so this was added for that. The local instance of store the index does not filter local addresses

scott-dn: > The local instance of store the index does not filter local addresses
But it can be a problem for the local gateway (prod is fine), as they got swarm addresses in the provider result list.

kckeiks: #158 we try all provider addresses until we run out of addresses, get an error, get a 404 or a valid response

scott-dn: Sound good. 👍

scott-dn (dismissed):

crates/ursa-index-provider/src/engine.rs
theBeardA: this depends on the design. the idea of the storetheindex was to advertise p2p address along with transport protocol(bitswap or graphsync) where peers can get data from. like it was one piece of content routing on the p2p(cache) node. But querying from the gateway, now we need server address. @b0xtch

kckeiks: Let me clarify. We pick one provider and try all of its advertised addresses (until we run out of addresses, get a 404 or any other response). 

Updated tests.

kckeiks: https://github.com/fleek-network/ursa/issues/187

scott-dn (approved):

PR #190, opened 2022-12-24
refactor(kad): start initial bootstrap from behaviour.rs sooner + improve logs

ozwaldorf:
start the bootstrap sooner, currently getting an error bootstrap timed out with no more bootstraps remaining. This seems to fix that getting 8+ bootstraps remaining after the first one instead of none. 

Manually tested + unit tested, there are still some timeouts but the bootstrap keeps going. While the bootstrap are returning some odd peer ids (`1AdivKqEdvpxsBm1yQjTRYbqk7MdfVjRmQAzxm752XQVZ6` vs our peer ids which always look like: `12D3KooWN2GSWDQwtRWsgiYXDgCHX5scrdHQPman3tBH2WtyCyK3`), it seems to be working. during the time it took to get to the final bootstrap response, the node had peer connection events for 30+ new peers. I think some timeouts are from peers that had id's switch and are no longer available at the /p2p/<id> address that might still persist in some routing tables

tldr; improves boostrapping consistency by starting it sooner, and new peers are flooding in.

### DIFF
--- a/crates/ursa-network/src/behaviour.rs
+++ b/crates/ursa-network/src/behaviour.rs
@@ -200,6 +200,16 @@ impl<P: StoreParams> Behaviour<P> {
             }
         }
 
+        if !config.bootstrapper && !config.bootstrap_nodes.is_empty() {
+            if let Err(e) = kad.bootstrap() {
+                warn!("Failed to bootstrap: {}", e);
+            } else {
+                info!("Bootstrapping into the network...");
+            }
+        } else {
+            warn!("Skipping bootstrap");
+        }
+
         Behaviour {
             ping,
             autonat,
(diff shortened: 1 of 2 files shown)


b0xtch:
Does this improve consistency or solve the issue

b0xtch (approved):

kckeiks:
This is interesting considering that the code that started the initial bootstrap is now only a few lines above. Does this change allow more time to avoid timeouts?

LGTM btw just curious.

kckeiks (approved):

ozwaldorf:
It's essentially restoring the location where the bootstrap is called to how it was before when kad was in `discovery.rs`, where creating the new behavior calls bootstrap() instead of later on after some time in `service.rs`

b0xtch (approved):

PR #192, opened 2022-12-24
feat: `make compose-build` + only build ursa binary in dockerfile

ozwaldorf:
## Why?

- Sequential docker builds are slower considering cargo builds all dependencies every time from scratch
- the dockerfile is building every single binary found in the project, instead of just the ursa bin specifically.
- no easy make helper to rebuild the docker-compose ursa image 

## What?

Adds [cargo-chef](https://github.com/LukeMathWalker/cargo-chef) caching to our docker file to take advantage of docker layers to not have to build all dependencies from scratch every time. This saves time by several minutes after the first build, so for node operators updating the deploy time is much sooner, as well as reducing build time for node development.

Also adds a new makescript `compose-build` to build ursa binary within the docker-compose context
 
## Demo?

First build:
image

Next builds with cached layers:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,44 +1,38 @@
-FROM rust:latest AS rust
+FROM rust:latest as builder
 
-# setup build environment
-WORKDIR /app
-RUN cargo install cargo-chef
-RUN cargo install cargo-strip
+WORKDIR /usr/src/app
+
 RUN apt-get update && apt-get install -y \
     clang \
     cmake \
     libclang-dev \
     protobuf-compiler
 
-FROM rust as planner
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    cargo install cargo-strip
 
 COPY . .
-RUN cargo chef prepare  \
-    --recipe-path recipe.json
 
-FROM rust as builder
-
-# build dependencies
-COPY --from=planner /app/recipe.json .
-RUN cargo chef cook \
-    --recipe-path recipe.json \
-    --bin ursa --release
-
-# build application
-COPY . .
 ENV RUST_BACKTRACE=1
-RUN cargo build --bin ursa --release && \
-    cargo strip && \
-    mv ./target/*/ursa .
+
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/usr/src/app/target \
+    cargo build --release --bin ursa \
+    && cargo strip \
+    && mv /usr/src/app/target/release/ursa /usr/src/app/
 
 FROM debian:bullseye-slim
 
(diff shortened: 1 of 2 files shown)


b0xtch:
Docker image is best to stay bare bones

Makefile
```diff
@@ -21,6 +21,9 @@ docker-build:
 docker-run:
 	docker run -p 4069:4069 -p 4070:4070 -p 6009:6009 -p 8070:8070 --name ursa-cli -it ursa
 
+compose-build:
```
b0xtch, line 24: Do we need build if we have up

ozwaldorf: up doesn't trigger rebuilds

ozwaldorf:
> Docker image is best to stay bare bones

The final image hasn't changed, it's just as bare (just the binary is copied from builder like before). just a new layer in the build process for dependencies specifically so that docker can cache that

b0xtch:
> > Docker image is best to stay bare bones
> 
> The final image hasn't changed, it's just as bare (just the binary is copied from builder like before). just a new layer in the build process for dependencies specifically so that docker can cache that

Which is what we already do. So we don’t currently need this

ozwaldorf:
Okay I can remove chef stuff, But the docker could still use the improvement of only building the ursa binary instead of every single binary in the workspace as well as the make script

b0xtch:
> Okay I can remove chef stuff, But the dockerfile could still use the improvement made of only building the ursa binary instead of every single binary in the workspace. As well as the make script addition

I think it does its job pretty well now. Any extra is sugar. Currently it builds it once and mounts cache and uses it for subsequent. Also when I meant bare bones as in how opinionated we should be on the image. This is just a bare min image just like we have a bare min compose, etc. less opinions the better

Dockerfile
```diff
@@ -35,4 +35,4 @@ COPY --from=builder /usr/src/app/ursa /usr/local/bin
 # run ursa node
 ENV RUST_LOG=info
 
-ENTRYPOINT ["ursa"]
```
b0xtch, line 38: What's the change here?

b0xtch (approved):

ozwaldorf:
with reverting all other than the single change to the dockerfile, build time is still improved by 30 seconds due to not building ursa-gateway, and ursa-tracker

Dockerfile
ozwaldorf: editor inserting newline at end of file i think

b0xtch (approved):

ozwaldorf:
gonna go ahead and merge, the failed test was a flakey in index provider and unrelated to the changes here 🙂

Issue #122, opened 2022-12-13
feat: gateway index workers


scott-dn:
Motivation:
- #121

How:
- jobs fetching results from indexers and propagating results to caches.

PR:
- #195

scott-dn:
solved at #195

PR #195, opened 2022-12-27
feat: gateway worker

scott-dn:
see #122


crates/ursa-gateway/src/config.rs
```diff
@@ -107,6 +113,9 @@ impl Default for GatewayConfig {
                 max_size: 200_000_000,  // 200MB
```
b0xtch, line 113: any plans to increase this?

crates/ursa-gateway/src/indexer/mod.rs
```diff
@@ -1,52 +1,182 @@
...
+        //   4.1 return block?
+        //   4.2 resolve?
+        //
+        // IMPROVEMENTS:
+        // 1. maintain N workers keep track of indexing data
+        // 2. cherry-pick closest node
+        // 3. cache TTL
+    }
+
+    pub async fn resolve_content(&self, addrs: Vec<SocketAddrV4>, cid: &str) -> Result<Vec<u8>> {
```
b0xtch, line 80: I am guessing this is just temp, as this function will be replaced and we should support streaming

b0xtch (approved):
Mint

crates/ursa-gateway/src/indexer/mod.rs
scott-dn: Yup 💯

crates/ursa-gateway/src/config.rs
scott-dn: I will create a ticket to track this issue status; thanks for reminding 🙏

crates/ursa-gateway/src/worker/cache.rs
```diff
@@ -0,0 +1,106 @@
...
+                key: String::from(k),
+            })?;
+            Ok(Arc::clone(data))
+        } else {
+            let (tx, rx) = oneshot::channel();
+            self.tx.send(WorkerCacheCommand::Fetch {
+                cid: String::from(k),
+                sender: tx,
+            })?;
+            let data = rx.await??;
```
kckeiks, line 88: @scott-dn I don't think this model of [send request to node] -> [read response and cache it] -> [send respond] is not going to work with stream piping the IO (client <-> node). Do you have some ideas?

scott-dn: Yeah, with cache, we don't have to deal with stream data at the moment (we might want to stream from cache).
But for the case of files larger than 1GB, we don't want to cache them; we should stream all the way instead.

kckeiks: How would we know the size before requesting content from the cache nodes or are you thinking that we will use `content-lenght`?

scott-dn: block info will do.
> https://github.com/fleek-network/ursa/pull/158#discussion_r1057999388

kckeiks: @theBeardA so what does the "block info" rpc method return?

theBeardA: we dont have "block info" method yet, what I meant in the comment was blockchain info and not info about data.
but if you want to know the size of the data, the root block have that information I think. so maybe we can use ursa_get_cid to get the root block and get the info from that @kckeiks

scott-dn: @theBeardA hmm, interesting; maybe cache node RPC can provide `block info`?
If that's too much, gateway can inspect the root block for the information.

theBeardA: I don't think it's going to be too much to add this rpc method on the cache node.
but isn't this something resolver should be doing? just a design choice.
lmk what you think @scott-dn @b0xtch @kckeiks

kckeiks: If we have to make a call to inspect a root block just to get size info about the content, that's another RTT.

kckeiks: We could inspect HTTP headers for things such as `Range` or`transfer-encoding: chunk` to indicate intent of download a large file then pipe the io.

scott-dn: > If we have to make a call to inspect a root block just to get size info about the content, that's another RTT.

It will be fine if we can have a cache to track content size history.


> We could inspect HTTP headers, from the client request, for things such as `Range` or`transfer-encoding: chunk` to indicate intent to download a large file then pipe the io.

Sounds good.

kckeiks: How about we include the file size in the metadata?

scott-dn: > How about we include the file size in the metadata?

isn't it another round trip to the indexer?

kckeiks: I don't think so. Aren't we currently calling the global indexer, given a cid? That returns the metadata. So instead of adding another rpc method or calling ursa_get_cid to get the file size, we could add the file size in the metadata. Thoughts?

scott-dn: > > isn't it another round trip to the indexer?
> 
> I don't think so. Aren't we currently calling the global indexer, given a cid? That returns the metadata. So instead of adding another rpc method or calling ursa_get_cid to get the file size, we could add the file size in the metadata. Thoughts?

sounds good to me.

kckeiks: That would be a lot easier then messing around with HTTP headers 😅

scott-dn: Agree 💯 we should have an issue ticket for this. 🙏

kckeiks: Here https://github.com/fleek-network/ursa/issues/206

Left some questions, notes, links from research.

kckeiks: https://github.com/fleek-network/ursa/issues/207

PR #209, opened 2022-12-29
feat: gateway streaming handler

scott-dn:
#198 

#### DEMO
```bash
$ curl http://0.0.0.0:4069/bafybeifyjj2bjhtxmp235vlfeeiy7sz6rzyx3lervfk3ap2nyn4rggqgei -v > origin
$ curl https://0.0.0.0:80/bafybeifyjj2bjhtxmp235vlfeeiy7sz6rzyx3lervfk3ap2nyn4rggqgei -k -v > gateway

$ openssl dgst -sha256 origin | awk '/.+$/{print "0x"$2}'
0x38b4a497f9ceea9d4fd024ee0d16b9989b71be2d4e8e10eaa67091d5db32f33e

$ openssl dgst -sha256 gateway | awk '/.+$/{print "0x"$2}'
0x38b4a497f9ceea9d4fd024ee0d16b9989b71be2d4e8e10eaa67091d5db32f33e
```

### DIFF
--- a/crates/ursa-gateway/src/server/route/api/v1/get.rs
+++ b/crates/ursa-gateway/src/server/route/api/v1/get.rs
@@ -1,28 +1,62 @@
 use std::{str::FromStr, sync::Arc};
 
-use axum::{extract::Path, response::IntoResponse, Extension, Json};
+use axum::{
+    body::StreamBody,
+    extract::Path,
+    http::{header, StatusCode},
+    response::{IntoResponse, Response},
+    Extension,
+    Json,
+};
 use cid::Cid;
-use hyper::StatusCode;
 use serde_json::{json, Value};
-use tokio::sync::RwLock;
+use tokio::{
+    io::{duplex, AsyncWriteExt},
+    sync::RwLock,
+    task::spawn,
+};
+use tokio_util::io::ReaderStream;
+use tracing::error;
 
-use super::super::super::super::model::HttpResponse;
-use crate::worker::cache::ServerCache;
+use crate::{server::model::HttpResponse, worker::cache::ServerCache};
 
 pub async fn get_block_handler<Cache: ServerCache>(
     Path(cid): Path<String>,
     Extension(cache): Extension<Arc<RwLock<Cache>>>,
-) -> impl IntoResponse {
+) -> Response {
     if Cid::from_str(&cid).is_err() {
         return error_handler(
             StatusCode::BAD_REQUEST,
             format!("Invalid cid string, cannot parse {cid} to CID"),
-        );
+        )
+        .into_response();
     };
 
     match cache.read().await.get_announce(&cid).await {
-        Ok(data) => (StatusCode::OK, Json(json!(data.as_ref()))),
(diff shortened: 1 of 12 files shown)


crates/ursa-gateway/src/worker/cache.rs
```diff
@@ -69,30 +93,80 @@ impl WorkerCache for Cache {
...
+                    body,
+                ) => body,
+                resp => {
+                    error!("Error requested provider: {resp:?}");
+                    bail!("Error requested provider");
+                }
+            };
+            let key = String::from(k); // move to [worker|writer] thread
+            let tx = self.tx.clone(); // move to [worker|writer] thread
+            spawn(async move {
```
kckeiks, line 144: Isn't this still buffering the response from the cache node in memory?

kckeiks: This is a streaming feature though so nvm. We can discuss further if we need piping.

kckeiks (approved):

crates/ursa-gateway/src/worker/cache.rs
scott-dn: yup, it's a streaming feature; we might need a pipeline solution one day.

crates/ursa-gateway/src/worker/cache.rs
```diff
@@ -69,30 +93,80 @@ impl WorkerCache for Cache {
...
+                            bytes.put(buf);
+                        }
+                        Err(e) => {
+                            error!("Failed to read stream for {e:?}");
+                            return;
+                        }
+                    }
+                }
+                if let Err(e) = tx.send(WorkerCacheCommand::InsertSync {
+                    key,
```
kckeiks, line 162: Does each chunk get a different key (cid)?

scott-dn: Yes, each will have diff cid.

crates/ursa-gateway/src/server/route/api/v1/get.rs
```diff
@@ -1,28 +1,47 @@
...
 use cid::Cid;
-use hyper::StatusCode;
 use serde_json::{json, Value};
 use tokio::sync::RwLock;
 
-use super::super::super::super::model::HttpResponse;
-use crate::worker::cache::ServerCache;
+use crate::{server::model::HttpResponse, worker::cache::ServerCache};
 
 pub async fn get_block_handler<Cache: ServerCache>(
```
b0xtch, line 16:
```suggestion
pub async fn get_car_handler<Cache: ServerCache>(
```

b0xtch (approved):
Mint, regarding the generic cache. Would it be better to create an impl for IPLD with support for`Tlrfu`. Food for thought only because this already support a generic `v` value. With deduplication, so pretty good. 

Example:

```rust
impl Tlrfu for IPLD {
   //
}
```

crates/ursa-gateway/src/server/route/api/v1/get.rs
scott-dn: Good catch 👍

kckeiks (approved):

PR #217, opened 2022-12-30
chore: dockerize gateway, reorg into /docker

b0xtch:
- [x] move all docker stuff under `/docker`
- [x] dockerize ursa-gateway
- [x] clean up infra and old setup
- [x] Add independent grafana dashboard for cache node


Makefile
```diff
@@ -19,7 +19,7 @@ docker-build:
 	docker build -t ursa -f ./Dockerfile .
 
 docker-run:
-	docker run -p 4069:4069 -p 4070:4070 -p 6009:6009 -p 8070:8070 --name ursa-cli -it ursa
+	docker run --name ursa-cli -it ursa
```
ozwaldorf, line 22: still need 6009 right?

infra/global-metrics/data/grafana/grafana.ini
```diff
@@ -7,6 +7,6 @@ default_home_dashboard_path = /etc/grafana/dashboards/ursa.json
 # allow anonymous users to view dashboards
 [auth.anonymous]
 enabled = true
-org_name = Main Org.
+org_name = Fleek Network.
```
ozwaldorf, line 10: need to setup this org as the main org as well separately

ozwaldorf (changes requested):
looking good, few comments. On moving the shared node grafana config, need to also update the docker compose for the tracker as it will bind mount the grafana configs to the necessary location.

Makefile
b0xtch: users who want that exposed should use the full-node impl

infra/global-metrics/data/grafana/grafana.ini
b0xtch: can we rename this or is this a hard requirement

ozwaldorf: Hard requirement to have it the default `Main Org.` or you'd need to add an overall definition for the group if changing. It's just a semantics thing tho so reverting should be fine, not too important

b0xtch: [49dcb9c](https://github.com/fleek-network/ursa/pull/217/commits/49dcb9c51a65334bd80e5d491bad20cf14fad7e0)

scott-dn (dismissed):

ozwaldorf:
> On moving the shared node grafana config, need to also update the docker compose for the tracker as it will bind mount the grafana configs to the necessary location.

Don't forget! I may also refactor this at a later time to separate them. the focus of global and local node dashboards i think are different (ie, global metrics includes geo stuff, but a local metrics/scrape wouldn't provide that)

scott-dn (approved):

ozwaldorf (approved):

PR #219, opened 2022-12-30
feat: nodes exchange bloom filters

matthias-wright:
**Motivation:**
Cache nodes use bloom filters to keep track of the set of objects stored in their cache. Nodes share their bloom filters with the network. Nodes will keep track of the bloom filters that were sent by other nodes. When a requested CID is not in a node's cache, a node can search through the bloom filters of other nodes to find the CID.

**Changes:**
* After receiving a PUT request, a node will update its bloom filter and publish it via gossipsub
* Counting bloom filters are used because they have a remove operation (useful for cache eviction)
* If a node receives a gossipsub message containing a bloom filter from another node, it will store the bloom filter in a map <del>(unless the sender is blacklisted)</del>

**Future work:**
* Bloom filters should be persisted to disk
* The map of bloom filters that each node stores should be kept up to date to avoid stale nodes

 The unit test `test_send_cache_summary` was coauthored together with @kckeiks.

### DIFF
--- a/crates/ursa-network/src/service.rs
+++ b/crates/ursa-network/src/service.rs
@@ -715,7 +715,7 @@ where
                         .request_response
                         .send_request(peer, UrsaExchangeRequest(RequestType::CacheRequest(cid)));
                 }
-                // update bloom filter and share it with the network
+                // update cache summary and share it with the network
                 self.cached_content.insert(&cid.to_bytes());
                 let topic = Topic::new(URSA_GLOBAL);
                 if let Ok(cache_summary) = self.cached_content.serialize() {
--- a/crates/ursa-network/src/service.rs
+++ b/crates/ursa-network/src/service.rs
@@ -679,12 +679,15 @@ where
                 println!("cosmos");
             }
             NetworkCommand::Put { cid, sender } => {
+                // replicate content
                 let swarm = self.swarm.behaviour_mut();
                 for peer in &self.peers {
                     swarm
                         .request_response
                         .send_request(peer, UrsaExchangeRequest(RequestType::CacheRequest(cid)));
                 }
+                // update bloom filter
+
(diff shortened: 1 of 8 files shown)


crates/ursa-network/src/service.rs
```diff
@@ -464,6 +479,21 @@ where
                 message_id,
                 message,
             } => {
+                if !self.peer_blacklist.contains(&propagation_source) {
```
kckeiks, line 482: what's propagation_source vs message.source?

kckeiks, line 482: What about a unit test for the `peer_blacklist`? Is that being populated atm?

kckeiks:
This is great! This is another good candidate for a testground test plan.

crates/ursa-network/src/service.rs
matthias-wright: No, it is not being populated right now. I considered it but I think that deciding when/why/for how long to blacklist a peer is worthy of a separate PR.

kckeiks: In that case, it would be cleaner to not include the blacklisting code until it is used.

crates/ursa-network/src/service.rs
```diff
@@ -675,16 +697,26 @@ where
                         )
                     }
                 }
-
-                println!("cosmos");
             }
             NetworkCommand::Put { cid, sender } => {
```
theBeardA, line 701: is this NetworkCommand::Put supposed to be called by put_car api?
if not, where should it be called?

b0xtch (approved):
Mint, we might keep the bloom filter ephemeral. Similarly to peers discovered.

kckeiks (approved):

Issue #207, opened 2022-12-29
feat: advertise file size in Metadata


kckeiks:

kckeiks:
#221

Issue #210, opened 2022-12-29
feat: gateway cache conditional


scott-dn:
only cache content <= 1GB by default (change via `config.toml`)
otherwise, stream directly from the cache node to users.

kckeiks:
Merged #221. Now working on updating gateway to make use of file size and stream directly.

PR #221, opened 2022-12-31
feat: advertise file size

kckeiks:
Part of #210 

The gateway needs to stream large files directly from cache node without using its internal caching. The gateway needs to know the file size to make this decision. This proposal piggybacks on metadata. 

Input:
> -rw-r--r--   1 acadia  staff   27044 Dec 18 16:50 basic.car

Command:
> curl https://0.0.0.0:80/bafybeifyjj2bjhtxmp235vlfeeiy7sz6rzyx3lervfk3ap2nyn4rggqgei 

Output:
```
2023-01-10T16:28:22.387896Z  INFO ursa_gateway::worker: Dispatch FetchAnnounce command with cid: "bafybeifyjj2bjhtxmp235vlfeeiy7sz6rzyx3lervfk3ap2nyn4rggqgei"
2023-01-10T16:28:22.390121Z  INFO ursa_gateway::resolver: File size received 27044
```

### DIFF
--- a/crates/ursa-index-provider/src/advertisement.rs
+++ b/crates/ursa-index-provider/src/advertisement.rs
@@ -22,6 +22,8 @@ struct Metadata {
     // Data is specific to the identified protocol, and provides data, or a
     // link to data, necessary for retrieval.
     Data: Vec<u8>,
+    // Size of the content.
+    Size: u64,
 }
 
 #[allow(non_snake_case)]
@@ -47,11 +49,18 @@ pub struct Advertisement {
     pub IsRm: bool,
 }
 impl Advertisement {
-    pub fn new(context_id: Vec<u8>, provider: PeerId, addresses: Vec<String>, is_rm: bool) -> Self {
+    pub fn new(
+        context_id: Vec<u8>,
+        provider: PeerId,
+        addresses: Vec<String>,
+        is_rm: bool,
+        content_size: u64,
+    ) -> Self {
         // prtocolid for bitswap
         let raw_metadata = Metadata {
             ProtocolID: 0x0900,
             Data: b"FleekNetwork".to_vec(),
+            Size: content_size,
         };
         let metadata = bincode::serialize(&raw_metadata).unwrap();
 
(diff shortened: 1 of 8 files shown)


crates/ursa-index-provider/src/advertisement.rs
```diff
@@ -22,6 +22,8 @@ struct Metadata {
     // Data is specific to the identified protocol, and provides data, or a
     // link to data, necessary for retrieval.
     Data: Vec<u8>,
```
kckeiks, line 24: If I'm reading the docs correctly, this is for protocol (bitswap, graphsync, etc) specific stuff but we are just adding this extra stuff (FleekNetwork). Do you think it will be ok to add this other data for the file size? @theBeardA

theBeardA: with bitswap, it is ok since data is empty. And I just realized that with graphsync, we might need to more information to Data field.
but on second thought, the data field is just vec<u8>, so we can have any kind of data encoded. and then when we retrieve, we should know what we need to do with this data and what is the expected result from decoding this data e.g GraphSyncStruct and other fields we are adding

kckeiks: So I'm thinking something like 
```rust
struct Metadata {
  ProtocolID: u128,
  Size: u64,
  Data: Vec<u8>,
}
```

Simple enough for now. We can then serialize/deserialize Data for protocol information. The indexer does not validate metadata right? If so, I think we will be OK to move forward then.

kckeiks: Another question I have is, `Size` makes sense if we are querying a file so `Size` is the file size but what if we are querying a block, do we currently support `put`ting a single block as apposed to an entire file?

kckeiks: cc @scott-dn @theBeardA

scott-dn: Yes, we can add that, but in order to increase the order, it's expected to follow as below:
image

kckeiks: @scott-dn what about `Size`? Can we expect to have `Size` for any CID? Considering that CID can be CID of root for a file or just a CID for some block (not the entire file).

scott-dn: I don't think there are any problems with that. We're able to know the size when storing a car file on the cache node.

kckeiks: Ok, cool 😊 I just wanted to make sure I understood the current use cases. I'll keep moving forward with this PR.

kckeiks:
Closes #207

PR #222, opened 2023-01-02
fix: switch murmur3 implementation

matthias-wright:
The `murmur3` implementation from [`fasthash`](https://github.com/flier/rust-fasthash) has some compilation issues, see [here](https://github.com/flier/rust-fasthash/issues/27) and [here](https://github.com/flier/rust-fasthash/issues/24). Hence switching to [`fastmurmur3`](https://crates.io/crates/fastmurmur3). Benchmarks of the two implementations are very similar.


b0xtch (approved):

kckeiks (approved):
👍

PR #223, opened 2023-01-02
chore(telemetry): set tracing tree log level and simplify default

ozwaldorf:
## Why

Currently the code is overcomplicated to load the log level. The preference should be: cli flag, then env var, then the default. Also, the log level provided is not respected by tracing tree, and tracing logs are printed due it not having a filter applied.

## What

- refactor/simplify the log level used
- add filter to tracing tree
- only use tracing tree for ursa bin to avoid duplicate logs

## Demo

image

### DIFF
--- a/crates/ursa-telemetry/src/lib.rs
+++ b/crates/ursa-telemetry/src/lib.rs
@@ -11,7 +11,7 @@ pub struct TelemetryConfig {
     /// Service name.
     pub name: String,
     /// Service log level.
-    pub log_level: Option<String>,
+    pub log_level: String,
     /// Service json log output.
     pub pretty_log: bool,
     /// Tokio console support.
@@ -28,7 +28,7 @@ impl TelemetryConfig {
     pub fn new(name: &str) -> Self {
         Self {
             name: name.to_string(),
-            log_level: None,
+            log_level: "INFO".to_string(),
             pretty_log: false,
             tokio_console: false,
             tree_tracer: false,
@@ -38,7 +38,7 @@ impl TelemetryConfig {
     }
 
     pub fn with_log_level(mut self, log_level: &str) -> Self {
-        self.log_level = Some(log_level.to_owned());
+        self.log_level = log_level.to_string();
         self
     }
 
@@ -68,14 +68,12 @@ impl TelemetryConfig {
     }
 
     pub fn init(self) -> anyhow::Result<()> {
-        let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| {
-            EnvFilter::new(self.log_level.unwrap_or_else(|| "INFO".to_string()))
-        });
-
         let mut tracing_layers = vec![];
 
         if self.pretty_log {
-            let log_subscriber = fmt::layer().pretty().with_filter(env_filter);
+            let log_subscriber = fmt::layer()
+                .pretty()
+                .with_filter(EnvFilter::new(&self.log_level));
             tracing_layers.push(log_subscriber.boxed());
         }
 
--- a/crates/ursa-telemetry/src/lib.rs
+++ b/crates/ursa-telemetry/src/lib.rs
@@ -1,4 +1,5 @@
 use serde::{Deserialize, Serialize};
+use std::env;
 use tracing_error::ErrorLayer;
 use tracing_subscriber::{
     fmt, layer::SubscriberExt, util::SubscriberInitExt, EnvFilter, Layer, Registry,
@@ -11,7 +12,7 @@ pub struct TelemetryConfig {
     /// Service name.
     pub name: String,
     /// Service log level.
-    pub log_level: String,
+    pub log_level: Option<String>,
     /// Service json log output.
     pub pretty_log: bool,
     /// Tokio console support.
@@ -28,7 +29,7 @@ impl TelemetryConfig {
     pub fn new(name: &str) -> Self {
         Self {
             name: name.to_string(),
-            log_level: "INFO".to_string(),
+            log_level: None,
             pretty_log: false,
             tokio_console: false,
             tree_tracer: false,
@@ -38,7 +39,7 @@ impl TelemetryConfig {
     }
 
     pub fn with_log_level(mut self, log_level: &str) -> Self {
-        self.log_level = log_level.to_string();
+        self.log_level = Some(log_level.to_string());
         self
     }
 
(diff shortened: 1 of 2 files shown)


crates/ursa-telemetry/src/lib.rs
```diff
@@ -11,7 +11,7 @@ pub struct TelemetryConfig {
     /// Service name.
     pub name: String,
     /// Service log level.
-    pub log_level: Option<String>,
+    pub log_level: String,
```
b0xtch, line 14: should stay optional, in case it is not provided it will default to info

crates/ursa-telemetry/src/lib.rs
```diff
@@ -28,7 +28,7 @@ impl TelemetryConfig {
     pub fn new(name: &str) -> Self {
         Self {
             name: name.to_string(),
-            log_level: None,
+            log_level: "INFO".to_string(),
```
b0xtch, line 31: this can then reverted back to None

crates/ursa-telemetry/src/lib.rs
```diff
@@ -38,7 +38,7 @@ impl TelemetryConfig {
     }
 
     pub fn with_log_level(mut self, log_level: &str) -> Self {
-        self.log_level = Some(log_level.to_owned());
+        self.log_level = log_level.to_string();
```
b0xtch, line 41: same here

crates/ursa-telemetry/src/lib.rs
ozwaldorf: sounds good

crates/ursa-telemetry/src/lib.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/223/commits/acc01e7fc57f79afcd4300e3172bb6b462b7fc1a

crates/ursa-telemetry/src/lib.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/223/commits/acc01e7fc57f79afcd4300e3172bb6b462b7fc1a

crates/ursa-telemetry/src/lib.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/223/commits/acc01e7fc57f79afcd4300e3172bb6b462b7fc1a

crates/ursa-telemetry/src/lib.rs
```diff
@@ -68,14 +69,16 @@ impl TelemetryConfig {
     }
 
     pub fn init(self) -> anyhow::Result<()> {
-        let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| {
-            EnvFilter::new(self.log_level.unwrap_or_else(|| "INFO".to_string()))
-        });
+        let log_level = self
+            .log_level
+            .unwrap_or_else(|| env::var("RUST_LOG").unwrap_or_else(|_| "INFO".to_string()));
```
b0xtch, line 74: What's the difference in this change?

crates/ursa-telemetry/src/lib.rs
```diff
@@ -68,14 +69,16 @@ impl TelemetryConfig {
...
 
         let mut tracing_layers = vec![];
 
         if self.pretty_log {
-            let log_subscriber = fmt::layer().pretty().with_filter(env_filter);
+            let log_subscriber = fmt::layer()
+                .pretty()
+                .with_filter(EnvFilter::new(&log_level));
```
b0xtch, line 81: Similarly here, you moved instantiating here?

crates/ursa-telemetry/src/lib.rs
ozwaldorf: easily clonable + need to instantiate possibly multiple times (as mentioned below) It also correctly uses the provided log level, then environment, whereas before it'd override provided value with the environment unexpectedly

crates/ursa-telemetry/src/lib.rs
ozwaldorf: yes env filter is not clonable

crates/ursa-telemetry/src/lib.rs
b0xtch: Override the default env what you mean? It get the default set env otherwise it sets. Prob don't need to clone it I think the only change required in the PR is lowering log levels in ursa bin

ozwaldorf: Well if you provide explicit log level, shouldn't be a need to load from env. For example ursa bin prefers the flag value over env and finally default. But before this change it'd get overridden regardless of the flag if env is set

b0xtch: What is a flag is not provided?

b0xtch: Which shouldn't be preferred and only be optional

ozwaldorf: if not provided it falls back to environment or the default

b0xtch: Okay so similar to the original code? Am I missing something?

ozwaldorf: no, with this change flow is:
bin (flag > env > default) -> telemetry uses (optional value > env > default )

vs before:
bin (same flag>enc>default) -> telemetry (env > provided value from bin > default)

b0xtch: Yes that seems accurate. Telemetry should take optional or fallback.

b0xtch: Env > optional > default

ozwaldorf: > Env > optional > default 

So env var precedent over cli flag override? we should switch the logic in ursa bin to match in that case (either provide some(level) or none and let telemetry handle env/default ). But imho i think the flag should override all not env

b0xtch: I think we should remove the flag it duplicated logic

b0xtch: Also if you look at it from first principals you are building telemetry to fit the need of bin while telemetry should fit the need of all. Example gateway maybe and doesn't have a flag. Food for thought

ozwaldorf: > Also if you look at it from first principals you are building telemetry to fit the need of bin while telemetry should fit the need of all. Example gateway maybe and doesn't have a flag. Food for thought 

Ironically, gateway does have a log flag and the addition of that flag to ursa bin was inspired from there. But i think the changes fits both use cases because with a flag/explicit level, can use `with_log_level`, otherwise, not using that function would just result in env or default loaded once by telemetry

b0xtch: Okay two birds with one stone. Now both don't need a flag. Also I think the gateway was before this new crate

Issue #214, opened 2022-12-29
feat: handle cache requests from other nodes


matthias-wright:
**Motivation:**
Each node is responsible for replicating its cached content onto its connected peers. After receiving a PUT request, a node will send a caching request to its connected peers (#193).

**TODO:**
- The cache requests have to be handled.
- For now, each node should accept a cache request (and store the corresponding content) unless the request comes from a blacklistet node.
- Accepting a cache request means getting the content (using bitswap for now) and storing it.

kckeiks:
Closed by #224.

PR #224, opened 2023-01-02
feat: Use graphsync to pull data

kckeiks:
This adds `rs-graphsync` to ursa service. This still needs some work and testing. I'm working on other tasks that will probably come in separate prs. In this pr we add minimal support to perform a graphsync pull request from peers.

 I ended up having to fork `rs-graphsync` and downgrade `libipld` because of dependency conflicts. Opened #239 to address it in the future.

- [x] ~Handle pending queries when pulling from multiple peers (maybe naive version then robust version in a follow up PR)~ Handle `CacheRequest` and pull content using graphsync #214.
- [x] Integrate `graphsync::Blockstore` trait with UrsaStore ~(follow up PR)~
- [x] Add unit test

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ cid = "0.8.6"
 clap = { version = "4.0.29", features = ["derive"] }
 console-subscriber = "0.1.8"
 ctrlc = "3.2.4"
-db = { package = "forest_db", version = "0.2.0", git = "https://github.com/theBeardA/forest-rocksdb", branch = "chore/upgrade-db", features = ["rocksdb"] }
+db = { package = "forest_db", version = "0.2.0", git = "https://github.com/kckeiks/forest-rocksdb", branch = "chore/derive-debug-rocksdb", features = ["rocksdb"] }
 dirs = "4"
 dotenv = "0.15.0"
 ipld_traversal = { git = "https://github.com/kckeiks/rs-graphsync.git", branch = "downgrade-cid" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ ctrlc = "3.2.4"
 db = { package = "forest_db", version = "0.2.0", git = "https://github.com/theBeardA/forest-rocksdb", branch = "chore/upgrade-db", features = ["rocksdb"] }
 dirs = "4"
 dotenv = "0.15.0"
-ipld_traversal = { git = "https://github.com/retrieval-markets-lab/rs-graphsync.git", rev = "0a53cf5" }
+ipld_traversal = { git = "https://github.com/kckeiks/rs-graphsync.git", branch = "downgrade-cid" }
 fastmurmur3 = "0.1.2"
 fnv = "1.0.7"
 futures = "0.3.25"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ cid = "0.8.6"
 clap = { version = "4.0.29", features = ["derive"] }
 console-subscriber = "0.1.8"
 ctrlc = "3.2.4"
-db = { package = "forest_db", version = "0.2.0", git = "https://github.com/kckeiks/forest-rocksdb", branch = "chore/derive-debug-rocksdb", features = ["rocksdb"] }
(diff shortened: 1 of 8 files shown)


b0xtch:
Looking good so far; other features and sure you are thinking about are the selectors and integration with our routing feature that @matthias-wright is currently on. ignore if that's the plan already

crates/ursa-network/src/behaviour.rs
```diff
@@ -15,6 +15,8 @@
 
 use anyhow::Result;
 use cid::Cid;
+use graphsync::GraphSync;
+use ipld_traversal::blockstore::MemoryBlockstore;
```
b0xtch, line 19: we should prob use our Ursa Store here instead of Mem

Cargo.toml
```diff
@@ -54,7 +54,7 @@ futures-util = "0.3.25"
 fvm_ipld_blockstore = "0.1.1"
 fvm_ipld_car = "0.6.0"
 fvm_ipld_encoding = "0.3.0"
-graphsync = { git = "https://github.com/retrieval-markets-lab/rs-graphsync.git", rev = "0a53cf5" }
+graphsync = { git = "https://github.com/kckeiks/rs-graphsync.git",  branch = "downgrade-cid" }
```
kckeiks, line 57: I was in dependency hell for a while because of `cid` and `libipld` dependencies. I ended up having to fork `rs-graphsync` and downgrade `libipld`. Opened #239 to address it in the future.

b0xtch:
https://github.com/theBeardA/forest-rocksdb/pull/2

Merged this PR

crates/ursa-network/src/behaviour.rs
```diff
@@ -94,13 +101,21 @@ pub struct Behaviour<P: StoreParams> {
 
     /// request/response protocol implementation for [`UrsaProtocol`]
     pub(crate) request_response: RequestResponse<UrsaExchangeCodec>,
+
+    /// Graphsync for efficiently exchanging data between blocks between peers.
+    pub(crate) graphsync: GraphSync<G>,
```
kckeiks, line 106: @b0xtch This is generic to support `Rocksdb` or `MemoryDB` or other implementer of `GSBlockstore`. Is this what you meant? Let me know if I misunderstood.

b0xtch: Yeah, the underlying store is the same. So we can abstract it to that layer and then instantiate. Similar to this https://github.com/fleek-network/ursa/blob/ab3e8e407428f84085496f25d9771cf98b79d3e8/crates/ursa-network/src/service.rs#L250

kckeiks: So I understand clearly. Are you suggesting wrapping `store` in an abstraction like `GraphsyncStorage` so we could have it like this in `Behaviour`?
```rust
pub(crate) graphsync: GraphsyncStorage,
```
This would still need us to parameterize and bound because both are parameterized as well`GraphsyncStorage<S: Blockstore>` and `BitswapStorage<S: Blockstore>`.

`libp2p_bitswap::Bitswap` doesn't maintain ownership of `store` so that is why we currently don't bound for `Blockstore`.

kckeiks: Or maybe you mean something like ?
```rust
struct Behaviour<S: Blockstore, P> {
...
    pub(crate) graphsync: GraphSync<GraphsyncStorage<S>>
```

b0xtch: The latter for sure

kckeiks: Updated. A nice upgrade to `rs-graphsync` would be for it to handle `store` like `bitswap` does - without mainting ownership. It would allow us to remove the bound.

kckeiks:
I updated this PR so that we use graphsync to pull content when we are asked by other nodes to cache content.

crates/ursa-network/src/behaviour.rs
b0xtch: Sounds good 👌

b0xtch (approved):
Mint

b0xtch (dismissed):

matthias-wright (dismissed):

matthias-wright (approved):

PR #233, opened 2023-01-03
feat: unify index provider and ursa rpc server

ozwaldorf:
## Why

Currently index provider is listening on a unique port. This can instead be under a single, unified server/port

## What

- index provider new router() method called and given to existing rpc server
- update tests in rpc service and index provider
- fix spelling error for `ProviderEngine::command_receiver()`
- use tilde for default paths in configs

## Demo

image


crates/ursa-index-provider/src/config.rs
```diff
@@ -1,15 +1,10 @@
-use dirs::home_dir;
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
-const DEFAULT_DB_PATH_STR: &str = ".ursa/data/index_provider_db";
+const DEFAULT_DB_PATH_STR: &str = "~/.ursa/data/index_provider_db";
```
kckeiks, line 4: Is this fine for windows users?

crates/ursa-rpc-service/src/server.rs
```diff
@@ -32,15 +32,20 @@ where
...
+    ) -> Result<()> {
         info!(
             "Server (rpc, http{}) starting up",
             if metrics.is_some() { " + metrics" } else { "" }
         );
 
         let http_address = SocketAddr::from(([0, 0, 0, 0], config.port));
 
-        let service = MultiplexService::new(self.http_app(metrics), self.rpc_app());
+        let service = MultiplexService::new(self.http_app(metrics), self.rpc_app(index_provider));
```
theBeardA, line 48: index_provider server should be an http endpoint, and not rpc

theBeardA (commented):
can be cleaned up as a part of this PR
https://github.com/fleek-network/ursa/blob/180585c03cd5673214a8eeb092f390b58b2d926f/infra/ursa/data/nginx/app.conf#L41-L61

crates/ursa-rpc-service/src/server.rs
ozwaldorf: perfect, will make this change :+1:

ozwaldorf: https://github.com/fleek-network/ursa/pull/233/commits/43e8e00c4a8b9f53b9bee70a86542b41b68d5ad3

theBeardA (commented):
once we have these server unified, we have conflicts with endpoints
the stream endpoint and indexer endpoint, they both points to `/:cid`.
we can't change how indexers query index-provider, so the obvious solution is to change our get/put endpoints with some something like `/ursa/v0/:cid`

ozwaldorf:
> once we have these server unified, we have conflicts with endpoints the stream endpoint and indexer endpoint, they both points to `/:cid`. we can't change how indexers query index-provider, so the obvious solution is to change our get/put endpoints with some something like `/ursa/v0/:cid`

This makes sense, let me see what parts need to change for this and make them

theBeardA (commented):
also a follow up to this would be to move domain to network config since we don't need specific sub-domain for provider, this can be a separate PR. I will create an issue for that

ozwaldorf:
> also a follow up to this would be to move domain to network config since we don't need specific sub-domain for provider, this can be a separate PR. I will create an issue for that

I think this makes sense to be included here

ozwaldorf:
> also a follow up to this would be to move domain to network config since we don't need specific sub-domain for provider, this can be a separate PR. I will create an issue for that

I think adding it to server_config would be better, or even just above all configs as a global value

theBeardA (approved):

theBeardA (approved):

theBeardA (approved):

kckeiks (approved):

docker/full-node/data/nginx/app.conf
```diff
@@ -38,32 +38,10 @@ server {
...
-
-      proxy_pass http://ursa:8070;
-    }
-}
-
 server {
     listen 443 ssl http2;
     listen [::]:443 ssl http2;
-    server_name ursa.earth www.ursa.earth;
+    server_name node.ursa.earth www.node.ursa.earth;
```
b0xtch, line 44: What’s node.ursa.earth?

crates/ursa-rpc-service/src/server.rs
```diff
@@ -56,9 +60,10 @@ where
             .layer(Extension(self.rpc_server.clone()))
     }
 
-    pub fn http_app(&self, metrics: Option<Router>) -> Router {
+    pub fn http_app(&self, index_provider: Router, metrics: Option<Router>) -> Router {
         Router::new()
             .merge(http::routes::network::init::<S>())
+            .merge(index_provider)
```
b0xtch, line 66: An improvement here would to merge using a Ismaili syntax as the other like “ http::routes::metrics::” something like that

docker/full-node/data/nginx/app.conf
ozwaldorf: just an example domain it'd be edited by the deployer as per readme.md

crates/ursa-rpc-service/src/server.rs
ozwaldorf: It'd need a bigger refactor where rpc-service would depend on index provider and metrics to do that. Vs just externally passing them in

PR #234, opened 2023-01-03
refactor: 💡 docker run bind mount ursa config default path

heldrida:
The following declares a bind mount from host path to container, which is used to persist the ursa config directory from default ursa home directory.


b0xtch:
We should encourage more usage of docker-compose if persistence is required

b0xtch:
@heldrida, whats the status of this, can we close?

PR #241, opened 2023-01-04
fix: 🐛 docker compose expose ports to host

heldrida:


infra/ursa/docker-compose.yml
```diff
@@ -70,12 +70,16 @@ services:
     user: "$UID:$GID"
 
   ursa:
+    # default image name which matches suggested docker run version for cache
+    image: ursa
     build:
       context: ../../.
       dockerfile: Dockerfile
     restart: on-failure
     ports:
```
b0xtch, line 79: These ports shouldn’t be exposed. Only 6009

heldrida: Ok, I can see that you have removed the ports to the host ( https://github.com/fleek-network/ursa/commit/c9dcab850001f7fc23468a251024255d1552ccca ).

I'll then remove the `make docker-run` from the docs. If not required in dev, might too just be removed to avoid confusion.

heldrida:
Removing the `make docker-run` from the docs. If not required in dev, might too just be removed to avoid confusion  ( https://github.com/fleek-network/ursa/commit/c9dcab850001f7fc23468a251024255d1552ccca ).

PR #242, opened 2023-01-04
feat: auto-upgraded configs using defaults + log the differences

### LINKED ISSUE #135: chore: programmatic update config
If a new field is added to config or the type of any field is changed in the config. The start of the node would error out since the current config stored in `.ursa/config.toml` file is not going to match the `UrsaConfig` structure in the newer version of code.
Add a warning and let user choose to overwrite the current config file with the defaults. this should interactive `yes/no`.

If the user has some customize config, they need to save that somewhere to before the file is overwritten with the new config structure and defaults.

ozwaldorf:
## Why

ursa bin errors if config is invalid

## What

- Provide deserialization defaults to automatically populate missing values
- Calculate unified diff on the toml we end up with after parsing
- If changes were made, log the diff and save the new config to file

## Demo

Example showing loading a config with some fields removed and old invalid fields added

image

We can also easily rename a field by modifying an attribute on a field like so:

```rust 
#[derive(Deserialize)]
struct ExampleConfig {
   #[serde(
       default = "default_example_fn",
       rename(deserialize = "old_name", deserialize = "new_name")
   )]
   new_name: String,
}
```

Which would result in the old name getting overwritten while preserving the value due to serializing into the new name

## Other notes

- This is currently implemented to be non-blocking and just a warning to avoid issues with automated/headless setups etc. Should it instead have a confirmation step (that's optionally skippable with a flag) that allows user to choose to accept or deny the changes?

- This will remove comments from the config file

---

closes #135

### DIFF
--- a/crates/ursa-index-provider/src/config.rs
+++ b/crates/ursa-index-provider/src/config.rs
@@ -1,24 +1,34 @@
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
-const DEFAULT_DB_PATH_STR: &str = "~/.ursa/data/index_provider_db";
-
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ProviderConfig {
-    /// a domain where provider is listening dns/test-node.provider.ursa.earth
+    /// a domain where the node is listening, eg. `dns/test-node.ursa.earth`
+    #[serde(default)]
     pub domain: String,
-    /// indexer url to point to e.g. https://dev.cid.contact
+    /// indexer url to point to, eg. https://dev.cid.contact
+    #[serde(default = "ProviderConfig::default_indexer_url")]
     pub indexer_url: String,
     /// database_path for index provider db
+    #[serde(default = "ProviderConfig::default_database_path")]
     pub database_path: PathBuf,
+}
+
+impl ProviderConfig {
+    fn default_database_path() -> PathBuf {
+        "~/.ursa/data/index_provider_db".into()
+    }
+    fn default_indexer_url() -> String {
+        "https://dev.cid.contact".to_string()
+    }
 }
 
 impl Default for ProviderConfig {
     fn default() -> Self {
         Self {
             domain: "".to_string(),
-            indexer_url: "https://dev.cid.contact".to_string(),
-            database_path: DEFAULT_DB_PATH_STR.into(),
+            indexer_url: Self::default_indexer_url(),
+            database_path: Self::default_database_path(),
         }
     }
 }
--- a/crates/ursa-network/src/config.rs
+++ b/crates/ursa-network/src/config.rs
@@ -2,65 +2,102 @@ use libp2p::Multiaddr;
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
-pub const DEFAULT_TRACKER_URL: &str = "https://tracker.ursa.earth/register";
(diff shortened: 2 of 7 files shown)


crates/ursa-network/src/config.rs
```diff
@@ -2,65 +2,102 @@ use libp2p::Multiaddr;
...
+    #[serde(default = "NetworkConfig::default_keystore_path")]
     pub keystore_path: PathBuf,
     /// Temporary HTTP tracker url. This is used for pre-consensus node registrations.
     /// Defaults to devnet tracker.
+    #[serde(default = "NetworkConfig::default_tracker")]
     pub tracker: String,
 }
 
+impl NetworkConfig {
+    fn default_mdns() -> bool {
```
b0xtch, line 46: Can we not implement a default trait much more readable here instead of a function returning a truth value

crates/ursa-index-provider/src/config.rs
```diff
@@ -1,24 +1,34 @@
...
+    /// indexer url to point to, eg. https://dev.cid.contact
+    #[serde(default = "ProviderConfig::default_indexer_url")]
     pub indexer_url: String,
     /// database_path for index provider db
+    #[serde(default = "ProviderConfig::default_database_path")]
     pub database_path: PathBuf,
 }
 
+impl ProviderConfig {
+    fn default_database_path() -> PathBuf {
```
b0xtch, line 18: Like why these functions returning a single value? Surely instantiating a map or something else would be more pragmatic and still serve the dynamic config feature

ozwaldorf: Serde derive supports functions only per field defaults

```
#[serde(default = "function")]
fieldname: String
```

Using `#[serde(default)]` alone results in Default::default() being called for the type specifically not the struct field containing the type

crates/ursa-network/src/config.rs
ozwaldorf: Default is implemented just under it, but we can't use individual fields from there in the derive macro

b0xtch: So a complexity was introduced? We can use defaults from a struct then tried to alter it by introducing other functions. I am not sure I follow the logic here. Additionally a value can used and as I mentioned a default trait can be created check here  https://serde.rs/attr-default.html

ozwaldorf: have a read of that document. You can only provide custom defaults through a function as mentioned in my other response

b0xtch: Have a read at our old code where we use defaults and we should stick to that

ozwaldorf: If i'm thinking of what you're saying correctly, he old code did use Option<T> on some fields and insert defaults after the fact, but that isn't on deserialization but rather usage and we'd need to make all current and future fields in the config struct Option<T> and fill them with Some(T) when loading the config, which requires unwraps or filling into a secondary struct to use throughout the codebase nicely. Whereas we can just call a defined function to fill the default value and directly use the fields without unwraps or additional handling

theBeardA (approved):

b0xtch (approved):

ozwaldorf:
Just leaving final comment: the flow required by serde is not the best looking, but due to lack of good alternatives, it's what we'll go with for now. Of course, if a cleaner refactor ever arises, it'd be much appreciated :pray:

PR #250, opened 2023-01-05
chore: run rust gh on crates

b0xtch:


.github/workflows/rust.yml
```diff
@@ -6,8 +6,12 @@ env:
 on:
   push:
     branches: [ "main" ]
+    paths:
+      - 'crates/**'
```
ozwaldorf, line 10: Can you include the workspace `Cargo.toml` since a dependency bump would be skipped?

ozwaldorf:
can we limit docker build as well?

.github/workflows/rust.yml
b0xtch: [2f8c7d4](https://github.com/fleek-network/ursa/pull/250/commits/2f8c7d4df565fdb7730a69a32feed5948e226a9b)

PR #268, opened 2023-01-09
feat: send cache summary using request response

matthias-wright:
## Why
At the moment, after receiving a PUT request, nodes will send their bloom filters to the network using gossipsub.
This will significantly increase network traffic. I still think that nodes should gossip their bloom filters. If neither a node A, nor its connected peers store the requested content, but some node B on the network has it, we want node A to know about node B with some probability. However, this should be done more carefully. For example, only gossipping bloom filters every N seconds to the network.

## What

- After receiving a PUT request, a node will no longer gossip its bloom filter to the network
- Instead, a node will send the bloom filter to its connected peers only using the request response behaviour
- If a node receives a bloom filter from a peer, it will store the bloom filter in a map (as before)
- If a peer disconnects, the bloom filter will be removed from the map

## Demo

See the unit test `test_send_cache_summary` .

## Checklist

### DIFF
--- a/crates/ursa-network/src/codec/protocol.rs
+++ b/crates/ursa-network/src/codec/protocol.rs
@@ -1,3 +1,4 @@
+use crate::utils::bloom_filter::CountingBloomFilter;
 use async_trait::async_trait;
 use cid::Cid;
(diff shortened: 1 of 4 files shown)


b0xtch (approved):
Mint

PR #274, opened 2023-01-10
feat: filter peers using cache summary for getbitswap

matthias-wright:
## Why

At the moment, `GetBitswap` will send requests to all connected peers, which can be inefficient. We therefore want to leverage the cache summaries (bloom filters) sent from the connected peers to filter out peers that definitely don't store the requested content. If there doesn't exist a bloom filter for a connected peer (I cannot think of a case where this happens atm), we still send out the request

## What

- Filter peers before sending getbitswap request using the bloom filters.

## Checklist

### DIFF
--- a/crates/ursa-network/src/service.rs
+++ b/crates/ursa-network/src/service.rs
@@ -757,10 +757,17 @@ where
                         self.response_channels.insert(cid, vec![sender]);
                     }
 
-                    let query = self
-                        .swarm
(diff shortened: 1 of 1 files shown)


matthias-wright:
I didn't create a unit test for this because all components used here are already unit tested (`test_bitswap_get()` and `test_send_cache_summary()`).

kckeiks (approved):

PR #278, opened 2023-01-11
feat: ipfs origin

### LINKED ISSUE #134: feat: ipfs gateway for origin
As a starting point for origin, we will cache miss into IPFS gateway if the content is not on the network.
there are ideas behind the origin field in the metadata but that can be tackled with another issue when we start integrating storage protocols

ozwaldorf:
## Why

Nodes should request content from origin on cache miss after bitswap queries.

Resolves #134 

## What

- Added OriginBehaviour allowing optimized origin requests
- Added service test and rpc tests

### Cli
image

### Tests
image

## Notes

- Concurrent requests for a cid from origin will be grouped like bitswap does
- currently using `CarReader::read_into` which is available on latest git for `fvm_ipld_car`. As soon as the release is available, we should switch back to crates.io

## Checklist


ozwaldorf:
btw: requested @b0xtch @kckeiks not sure why it says removed botch ^

ozwaldorf:
Noticed an issue, marking draft for now

crates/ursa-network/src/origin.rs
```diff
@@ -0,0 +1,178 @@
...
+pub struct OriginBehavior<S> {
+    store: Arc<UrsaStore<S>>,
+    query_count: QueryId,
+    pending: FnvHashMap<QueryId, PendingQuery>,
+    client: Arc<Client>,
+    config: OriginConfig,
+    event_queue: VecDeque<OriginEvent>,
+}
+
+impl<S> OriginBehavior<S>
```
kckeiks, line 85: Why do we need a behaviour to fetch data from the origin if we're just going to use `surf::Client`, wouldn't it be more simple to just make that call in `check_blockstore`?

ozwaldorf: The goal is to share successful responses for concurrent requests for cids from both bitswap and origin

Also, as we add more origins the requirements to get from them will differ and extend further than simple http calls

And finally to note I may switch surf client -> hyper client here, lmk if you have thoughts on it

crates/ursa-network/src/behaviour.rs
```diff
@@ -116,6 +121,7 @@ where
     pub fn new<B: BitswapStore<Params = P>>(
         keypair: &Keypair,
         config: &NetworkConfig,
+        origin_config: &OriginConfig,
```
b0xtch, line 124: This is replicated behaviour, you already have access to origin_config through `NetworkConfig`

crates/ursa-network/src/origin.rs
```diff
@@ -0,0 +1,178 @@
...
+pub enum Origin {
+    Ipfs,
+    Arweave,
+    Filecoin,
+}
+
+impl Origin {
+    pub fn iter() -> Iter<'static, Origin> {
+        [Origin::Ipfs].iter()
+    }
```
b0xtch, line 42: Is this used?

crates/ursa-network/src/behaviour.rs
ozwaldorf: It's added as a top level config, not under network config. I could switch origin config under network config but I had thought that it made more readable as it's own instead of nested

crates/ursa-network/src/origin.rs
ozwaldorf: Good catch! I initially had an implementation where providing origin was optional, and if unprovided we could easily iterate over the list of supported origins and try them one by one using this, but instead went with a specific origin per request. Will remove this :+1:

crates/ursa-network/src/behaviour.rs
b0xtch: You still have crate access to it. `config::OriginConfig`

crates/ursa-network/src/service.rs
```diff
@@ -751,11 +777,7 @@ where
                     )))
                         .map_err(|_| anyhow!("Failed to get a bitswap block!"));
                 } else {
-                    if let Some(chans) = self.response_channels.get_mut(&cid) {
-                        chans.push(sender);
-                    } else {
-                        self.response_channels.insert(cid, vec![sender]);
-                    }
+                    self.response_channels.push_bitswap(cid, sender);
```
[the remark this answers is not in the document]
ozwaldorf: It utilizes [HashMap's Entry API](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) to initialize with an empty vector if nonexistent

crates/ursa-network/src/origin.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/c329178e1f5d47d673bdcf7a65187e8fcd02bd66

crates/ursa-network/src/behaviour.rs
ozwaldorf: Not sure what that solves, NetworkConfig and OriginConfig are separate entities passed in from the ursa bin

ozwaldorf: Not sure exactly what you mean, Origin and Network Configs are separate entities passed from the ursa binary

crates/ursa-network/src/service.rs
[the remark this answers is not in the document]
b0xtch: does it cover this case from the original code, when its empty not push an empty vector but a vector of sender
```
self.response_channels.insert(cid, vec![sender]);
```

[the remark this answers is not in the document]
b0xtch: Also just checked the api, could link me to the `initialize with an empty vector if nonexistent for such cases`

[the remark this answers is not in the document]
ozwaldorf: hmm, i suppose it differs slightly due to using or_default. It can be improved by or_insert, I'll amend with the optimization

crates/ursa-network/src/behaviour.rs
b0xtch: OriginConfig can be part of network config as a param or something contained.

crates/ursa-network/src/origin.rs
kckeiks: Oh I see. You don't need a behaviour to communicate to Bitswap about an event though, you can use a channel for that. 

> Also, as we add more implementations for origins, the requirements to get from them will differ and extend further than just simple http calls

I recommend that we get a better idea of requirements before writing code for it. You can avoid a lot of work if you know what you need.

I like the idea of concurrent requests but we should be cautious of optimizing too early in development before having some data and being able to measure the trade off such as bandwidth and latency.

crates/ursa-rpc-service/src/api.rs
```diff
@@ -103,41 +104,41 @@ impl<S> NetworkInterface for NodeNetworkInterface<S>
...
+    async fn check_blockstore(&self, cid: Cid) -> Result<()> {
         if !self.store.blockstore().has(&cid)? {
             info!("Requesting block with the cid {cid:?}");
             let (sender, receiver) = oneshot::channel();
             let request = NetworkCommand::GetBitswap { cid, sender };
 
             // use network sender to send command
             self.network_send.send(request)?;
+
             if let Err(e) = receiver.await? {
```
kckeiks, line 116: It doesn't look like it's happening concurrently if we're waiting on this receiver? Am I reading this right?

ozwaldorf: All open receivers here are triggered by any successful response for a given cid be it bitswap or origin

crates/ursa-network/src/service.rs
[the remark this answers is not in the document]
ozwaldorf: > Also just checked the api, could link me to the `initialize with an empty vector if nonexistent for such cases`

https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_default

crates/ursa-rpc-service/src/api.rs
kckeiks: But we havent made the origin request yet at this point, no? if this returns an error, it sends the GetOriging command, or is the GET origin request triggered somewhere before this line?

ozwaldorf: oh, i understand what you're saying now. Basically flow is this:
1. attempt bitswap
2. bitswap failed attempt origin

But concurrent users requesting the same content would resolve from a single request no matter what step they are at

ozwaldorf: It does bring up a point though, where we could request bitswap and origin at the same time and use whichever responds first

kckeiks: That's what I thought you meant with concurrent requests. 

> The goal is to share successful responses for concurrent requests for cids from both bitswap and origin, so that latency is reduced for new content, as well as retain the exact same flow of using bitswap or the origin'

kckeiks: > But concurrent users requesting the same content would resolve from a single request no matter what step they are at

In that case a new behavior isn't really necessary.

ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/11e9610901aad513e46cbede8f05907d30ba1363

crates/ursa-network/src/behaviour.rs
ozwaldorf: refactored implementation into rpc and moved it to `[server_config.origin]`
https://github.com/fleek-network/ursa/pull/278/commits/11e9610901aad513e46cbede8f05907d30ba1363

crates/ursa-network/src/origin.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/11e9610901aad513e46cbede8f05907d30ba1363

crates/ursa-rpc-service/src/api.rs
```diff
@@ -79,65 +91,188 @@ pub trait NetworkInterface: Sync + Send + 'static {
...
+                "https://{}/ipfs/{root_cid}",
+                self.origin_config.ipfs_gateway
+            )
+            .parse()?,
+        )
+        .header("Accept", "application/vnd.ipld.car")
+        .build();
+
+        let store = self.store.db.clone();
+        spawn(async move {
```
b0xtch, line 185:
```suggestion
        task::spawn(async move {
```

crates/ursa-rpc-service/src/api.rs
```diff
@@ -79,65 +91,188 @@ pub trait NetworkInterface: Sync + Send + 'static {
...
+            let res = async {
+                let res = client.send(req).await.map_err(|e| {
+                    format!("Error getting content for cid {root_cid} from origin: {e}")
+                })?;
+
+                let mut car = CarReader::new(res).await.map_err(|e| {
+                    format!("Error reading car file for cid {root_cid} from origin: {e}")
+                })?;
+
+                if car.header.roots.iter().any(|cid| *cid == root_cid) {
```
b0xtch, line 196: why aren't we using `load_car` directly?

crates/ursa-rpc-service/src/api.rs
```diff
@@ -79,65 +91,188 @@ pub trait NetworkInterface: Sync + Send + 'static {
...
+            client: Arc::new(Client::new()),
+        }
+    }
+
+    /// Fetch content from the network
+    async fn get_network(&self, root_cid: Cid) -> Result<()> {
+        info!("Fetching cid {root_cid} from network");
+        let (send, recv) = oneshot::channel();
+
+        // Dispatch bloom filtered bitswap request
```
b0xtch, line 143: verbose

crates/ursa-rpc-service/src/api.rs
b0xtch: also could we use the internal `put_file` method

ozwaldorf: I wanted to verify the cid is present in headers before caching the content

crates/ursa-rpc-service/src/api.rs
ozwaldorf: Why the verbosity? there's only one spawn type used in the file

crates/ursa-rpc-service/src/api.rs
b0xtch: How does that prevent from using load car directly tho

crates/ursa-rpc-service/src/api.rs
b0xtch: Because the entire project uses that verbosity and it's not verbose similar to your comments above. This is a task spawn as opposed to something else that can be spawned.

ozwaldorf: Sounds good just curious

crates/ursa-rpc-service/src/api.rs
ozwaldorf: I tried using CarReader and passing load_car(store, car.reader) but for some reason it's erroring when the CarReaders are nested

crates/ursa-rpc-service/src/api.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/2863d7d5d1be45e473db00c728f6e6e6339b2fd2

crates/ursa-rpc-service/src/api.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/2863d7d5d1be45e473db00c728f6e6e6339b2fd2

crates/ursa-rpc-service/src/api.rs
b0xtch: Interesting, you don't think using existing code like `put_file` or `put_car` is a good option. I am wondering why?

ozwaldorf: We have the data, owned and in memory already, why write to disk -> read from disk in put_file -> write to store in load_car, and finally read from store/traverse dag to return the data from `get`/`get_data`. And that's not even including if we want to ensure the root cid is present

The flow now is parse car -> verify root is present with CarReader -> insert directly into blockstore -> final read/traveral to return the data

ozwaldorf: In terms of put_car, maybe that could be used but I still think we should check to ensure the root cid is present

ozwaldorf: actually looking into it the previous implementation on current main is not properly handling this either for bitswap when cid is missing

b0xtch: Put file calls put car which in turn calls load car. 

So if you are using load_car then put_car is the same

b0xtch: We have to store the content from origin right? Because it's a cache miss

crates/ursa-rpc-service/src/api.rs
```diff
@@ -79,65 +91,187 @@ pub trait NetworkInterface: Sync + Send + 'static {
...
+            pending_requests: Arc::new(RwLock::new(HashMap::new())),
+            client: Arc::new(Client::new()),
+        }
+    }
+
+    /// Fetch content from the network
+    async fn get_network(&self, root_cid: Cid) -> Result<()> {
+        info!("Fetching cid {root_cid} from network");
+        let (send, recv) = oneshot::channel();
+
```
kckeiks, line 142:
```suggestion

```
We could reduce the white space since it's a small method.

crates/ursa-rpc-service/src/api.rs
kckeiks: > ... we still need to emit the provider and network command to trigger advertisement and content replication

Yes, and that's important for indexing. Another data point that is important to the gateway is the size of the file so we need to handle provide that. Please see #221.

crates/ursa-rpc-service/src/config.rs
```diff
@@ -22,6 +24,27 @@ impl Default for ServerConfig {
         Self {
             port: Self::default_port(),
             addr: Self::default_addr(),
+            origin: Default::default(),
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+pub struct OriginConfig {
```
kckeiks, line 33: Since this only holds one type of data, it would be less code to simply add it in ServerConfig.

crates/ursa-rpc-service/src/api.rs
kckeiks: cc @scott-dn

theBeardA (commented):
just a general comment,
the flow I understood is cache miss into origin -> get the data -> store the data into the node's blockstore -> stream it to back to client.
since the data fetched is new for the network, I think we should call content routing and indexer commands to let the content routing do it's thing when next time that data is requested

crates/ursa-rpc-service/src/config.rs
ozwaldorf: Sure, although we may need many fields for different origins, not sure if they'd be nice all top level in server config instead of nested

crates/ursa-rpc-service/src/api.rs
```diff
@@ -79,68 +88,228 @@ pub trait NetworkInterface: Sync + Send + 'static {
...
+        rx.recv()
+            .await
+            .ok_or_else(|| anyhow!("Failed to receive status from channel"))?
     }
 
-    async fn get_data(&self, root_cid: Cid) -> Result<Vec<(Cid, Vec<u8>)>> {
-        if !self.store.blockstore().has(&root_cid)? {
+    /// Trigger the network and provider to start providing the content id.
+    fn provide_cid(&self, cid: Cid, size: u64) -> JoinHandle<Result<()>> {
```
b0xtch, line 246: why couldn't we use `put_car` or `put_file`?

crates/ursa-network/Cargo.toml
```diff
@@ -36,6 +36,7 @@ tokio.workspace = true
 tracing.workspace = true
 ursa-metrics = { path = "../ursa-metrics" }
 ursa-store = { path = "../ursa-store" }
+void = "1.0.2"
```
b0xtch, line 39: is this used anywhere?

b0xtch: if it is, should be in the top level cargo

crates/ursa-rpc-service/src/api.rs
```diff
@@ -79,68 +88,228 @@ pub trait NetworkInterface: Sync + Send + 'static {
...
     pub network_send: Sender<NetworkCommand>,
     pub provider_send: Sender<ProviderCommand>,
+    pending_requests: PendingRequests,
+    client: Arc<Client>,
+    origin_config: OriginConfig,
 }
 
-#[async_trait]
-impl<S> NetworkInterface for NodeNetworkInterface<S>
+impl<S> NodeNetworkInterface<S>
```
b0xtch, line 115: Why was the async trait removed and `NetworkInterface` impl as well

ozwaldorf: it's not it's just the diff being weird. It's the internal implementation added above the trait implementation

crates/ursa-network/Cargo.toml
ozwaldorf: oop need to revert, thanks

crates/ursa-rpc-service/src/api.rs
```diff
@@ -79,68 +88,228 @@ pub trait NetworkInterface: Sync + Send + 'static {
...
+            provider_send,
+            origin_config,
+            pending_requests: Arc::new(RwLock::new(HashMap::new())),
+            client: Arc::new(Client::new()),
+        }
+    }
+
+    /// Check blockstore for content. On cache miss, attempt to fetch and start providing it
+    /// Returns Ok(true) if the block was found in the blockstore, and Ok(false) if the content was fetched
+    async fn check_blockstore(&self, cid: Cid) -> Result<bool> {
```
b0xtch, line 137: should we `check_blocstore` have one job instead of a nested `get_network`. Like what if I just want to check if a block exists

ozwaldorf: Maybe a better name for this fn would be better, like `check_or_fetch` ?

crates/ursa-rpc-service/src/api.rs
ozwaldorf: can't move self

crates/ursa-rpc-service/src/api.rs
```diff
@@ -317,3 +456,27 @@ where
         Pin::new(&mut self.reader).poll_read(cx, buf)
     }
 }
+
+pub async fn load_car_checked<R, B>(s: &B, reader: R, root_cid: Cid) -> Result<Vec<Cid>>
```
b0xtch, line 460: the latter half is still a copy of the `load_car`. we which prob call `load_car`

ozwaldorf: how do you propose the former half be done then? without copying data

crates/ursa-rpc-service/src/api.rs
```diff
@@ -216,39 +386,8 @@ where
...
-            error!("there was an error while sending provider command {e}");
-            return Ok(cids);
-        }
-        match receiver.await {
-            Ok(_) => Ok(cids),
-            Err(e) => Err(anyhow!(format!(
-                "The PUT failed, please check server logs {e:?}"
-            ))),
-        }
+        self.provide_cid(root_cid, size).await?.map(|_| cids)
```
b0xtch, line 390: here we spawn a nested task. We should prob revert this and just use `put_car`.

crates/ursa-rpc-service/src/api.rs
b0xtch: we should use `load_car` instead of copying and adding an if statement before the body. Can't async move self? Can you elaborate? I am assuming I can understand correctly. it is because you are spawing a unnecessary task

crates/ursa-rpc-service/src/api.rs
ozwaldorf: yes and immediately awaited, just a layer of indirection. need the spawn so that network and provider notification is non blocking, so that get/get_data responds without extra delay

crates/ursa-rpc-service/src/api.rs
b0xtch: I am trying to figure out if I follow. This the load car impl, https://github.com/filecoin-project/ref-fvm/blob/3baa50edf5c15d2645a7dc3cfbab718831397646/ipld/car/src/lib.rs#L155

and you just copy-paste this here:

```rust
    let mut buf = Vec::with_capacity(100);
    while let Some(block) = car_reader.next_block().await? {
        buf.push((block.cid, block.data));
        if buf.len() > 1000 {
            s.put_many_keyed(buf.iter().map(|(k, v)| (*k, v)))
                .map_err(|e| Error::Other(e.to_string()))?;
            buf.clear();
        }
    }
    s.put_many_keyed(buf.iter().map(|(k, v)| (*k, v)))
        .map_err(|e| Error::Other(e.to_string()))?;
    Ok(car_reader.header.roots)
```

crates/ursa-rpc-service/src/api.rs
ozwaldorf: If the delay from provider and network commands are negligible, then I can remove the task and use blocking

crates/ursa-rpc-service/src/api.rs
b0xtch: but it is not, it does block here and has to wait for the task. We have the server task which handles each of the commands separately, then you are spawning another task.

crates/ursa-rpc-service/src/api.rs
ozwaldorf: Both `get` and `get_data` had the same code that is now check_blockstore, I put it in a function so as to DRY

crates/ursa-rpc-service/src/api.rs
b0xtch: https://docs.rs/axum/latest/axum/ for reference

crates/ursa-rpc-service/src/api.rs
b0xtch: It doesn't block, we are spawning a task for every action as. Axum for reference https://docs.rs/axum/latest/axum/

crates/ursa-rpc-service/src/api.rs
b0xtch: Like did you move the block around is that why the diff is being weird?

crates/ursa-rpc-service/src/api.rs
b0xtch: I see that now, hard to see when block of code are just moved around arbitrarily

crates/ursa-rpc-service/src/api.rs
ozwaldorf: CarReader consumes the async reader to parse the headers, so I can't get the header and then pass the inner reader from CarReader to load_car, because it'll be missing headers and fail due to cbor. As a workaround this method performs the root cid check and the same logic from load car with a single CarReader, since nesting them would error. Of course the other solution would be to clone the data and parse twice but that's not nearly as efficient. 

I think a good improvement to the fvm_ipld_car crate would be to provide an implementation of load_car that when passed a CarReader it would correctly reuse headers and blocks

b0xtch: we can open PR there

b0xtch: Just reading over it now i don't this is a necessary step 

```
    // verify root cid is present in the car header
    if !car.header.roots.contains(&root_cid) {
        return Err(anyhow!("Root CID not present in car header"));
    }
```

Using the normal `load_car` vs `load_car_unchecked` should check the cids according to docs I linked above

crates/ursa-rpc-service/src/api.rs
ozwaldorf: no nothing was moved just inserted in between

crates/ursa-rpc-service/src/api.rs
b0xtch: This is not the cache miss handler so it shouldn't care about that function

b0xtch: you should be able to spawn this "blocking" task

crates/ursa-rpc-service/src/api.rs
b0xtch: for better readability which maintains structure when the structure is not required to change. makes reviewing hard

crates/ursa-rpc-service/src/api.rs
ozwaldorf: nothing was moved around it was put into an internal implementation because not used externally as recommended by https://github.com/fleek-network/ursa/pull/278#discussion_r1068751470

b0xtch: You mentioned it was inserted in between so it was moved

crates/ursa-rpc-service/src/api.rs
ozwaldorf: While the origin most likely respond with a valid CAR, i think there's a chance in a malicious scenario that the car could be incorrect, and we'd store data we don't care about. Do we want to 100% trust that origin gateways are serving car requested?

crates/ursa-rpc-service/src/api.rs
ozwaldorf:
```
axum -> request1
  |
  -> request2 -> check/fetch to store -> success response
                      \-> put command task
```

I can make this block so that
```
request2 -> check/fetch store -> put command if needed -> response to end user
```

But I think the former is better so that TTFB is lower

b0xtch: You have misrepresented how it currently works. It doesn't block each request. I recommended going over the code and Axum again

crates/ursa-rpc-service/src/api.rs
ozwaldorf: I can move it for you

b0xtch: I think in general not just for me viewing pleasure. For everyone's

crates/ursa-rpc-service/src/api.rs
b0xtch: There is not trust assumptions here. It's a car file

b0xtch: https://ipld.io/specs/transport/car/

Another good reference.

crates/ursa-rpc-service/src/api.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/189110264aa4e4398f5b0b5eafa5dec0cb61a20e

crates/ursa-rpc-service/src/api.rs
ozwaldorf:
```
match cid.hash().code() {
    0x0 => {
        if cid.hash().digest() != data {
            return Err(Error::InvalidFile(
                "CAR has an identity CID that doesn't match the corresponding data"
                    .into(),
            ));
        }
    }
    code => {
        let code = multihash::Code::try_from(code)?;
        let actual = Cid::new_v1(cid.codec(), code.digest(&data));
        if actual != cid {
            return Err(Error::InvalidFile(format!(
                "CAR has an incorrect CID: expected {}, found {}",
                cid, actual,
            )));
        }
    }
}
```
This is the verification, It's only making sure the CAR is valid to itself, not to the request we made to the gateway. Invalid CAR would be rejected here, but we still should make sure that the valid car we get back is for the root cid we intended to receive specifically, and not some arbitrary valid CAR

crates/ursa-rpc-service/src/api.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/fbc8f99c5acc0fce42e47de0f407faea4f53ff37

crates/ursa-rpc-service/src/api.rs
ozwaldorf: Ok i removed the task https://github.com/fleek-network/ursa/pull/278/commits/fbc8f99c5acc0fce42e47de0f407faea4f53ff37

crates/ursa-rpc-service/src/api.rs
ozwaldorf: resolving this, latest convo here on the extra verification https://github.com/fleek-network/ursa/pull/278#discussion_r1069939919

and provider and network is correct now

crates/ursa-rpc-service/src/api.rs
```diff
@@ -97,51 +108,43 @@ where
...
-                    e
-                ));
+    async fn get(&self, cid: Cid) -> Result<Vec<u8>> {
+        let new = self.check_blockstore(cid).await?;
+        let content =
+            self.store.blockstore().get(&cid)?.ok_or_else(|| {
+                anyhow!("content was fetched but could not be found in blockstore")
+            })?;
+        if new {
+            let size = content.len() as u64;
```
kckeiks, line 128: Doesn't `self.store.blockstore().get` return just one block? if so `content.len()` would be the length of just one block instead of the length/size of the file?

crates/ursa-rpc-service/src/api.rs
b0xtch: I need to follow how to check the root. You can get the root of the car file after you load it. This seems redundant. We can just go with using `load_car`

> The roots array must contain one or more CIDs, each of which should be present somewhere in the remainder of the CAR.

crates/ursa-rpc-service/src/api.rs
```diff
@@ -97,51 +108,43 @@ where
...
-                return Err(anyhow!(
-                    "The bitswap failed, please check server logs {:?}",
-                    e
-                ));
+        let new = self.check_blockstore(root_cid).await?;
+        let dag = self.store.dag_traversal(&root_cid)?;
+        if new {
+            // note: this is blocking, so we need to ensure that Provider and Network Put
+            // do not take excessive time to initiate to avoid delaying response time
+            let size = dag.iter().map(|(_, c)| c.len() as u64).sum();
```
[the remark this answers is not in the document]
ozwaldorf: Definitely agree, let me look into what the gateways send for headers. I wasn't sure if it was the content size or the full size of the car file including the headers etc.

For content fetched through bitswap though, it might be a little harder. I think we'd need to construct a car and calculate a total as bitswap wouldn't be sharing the car, just the blocks themselves right?

crates/ursa-rpc-service/src/api.rs
ozwaldorf: That's correct. I'm not sure how much of a use this method is though, as it's not currently used anywhere in the codebase. See https://github.com/fleek-network/ursa/pull/278#discussion_r1071471095 for my response on fixing the size calculation

crates/ursa-rpc-service/src/api.rs
b0xtch: Unresolved this, the original comment wasn't addressed

crates/ursa-rpc-service/src/api.rs
```diff
@@ -97,51 +108,43 @@ where
...
-                    "The bitswap failed, please check server logs {:?}",
-                    e
-                ));
+    async fn get(&self, cid: Cid) -> Result<Vec<u8>> {
+        let new = self.check_blockstore(cid).await?;
+        let content =
+            self.store.blockstore().get(&cid)?.ok_or_else(|| {
+                anyhow!("content was fetched but could not be found in blockstore")
+            })?;
+        if new {
```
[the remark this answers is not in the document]
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/b413770804dcd4fa6be135b4495340255af945b3 Is this suitable?

crates/ursa-rpc-service/src/api.rs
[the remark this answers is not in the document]
b0xtch: Yeah, we can add this func to our store crate, `size(cid: CId)`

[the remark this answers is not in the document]
kckeiks: > content size or the full size of the car file including the headers etc

I was referring to the size of the data the user wants. Whether we should include size of car headers, I'm not sure. I guess one thing to look out for is whether the origins provide size including headers, etc., then we could just provide that? Definitely some unknowns here so would be good to document afterwards so we know what our software is doing.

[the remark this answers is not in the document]
kckeiks: Whatever the result of the research is, let's make use it is consistent with https://github.com/fleek-network/ursa/pull/221. If work in #221 needs to be updated, we can track it in a separate issue.

crates/ursa-rpc-service/src/api.rs
ozwaldorf: I'll update the method with a better name, to clearer communicate what it's purpose is

> what if I just want to check if a block exists

Just using `store.blockstore().has(cid)` should suffice for that usecase

crates/ursa-rpc-service/src/api.rs
ozwaldorf: If we get the root after it's been loaded and it's missing, should anything be done about the arbitrary content just loaded?

crates/ursa-rpc-service/src/api.rs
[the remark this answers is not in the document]
ozwaldorf: Should be resolved with this one: 
https://github.com/fleek-network/ursa/pull/278/commits/ebb750854c81e7ec038c6e740b0cc8db0d4a3085

crates/ursa-rpc-service/src/api.rs
```diff
@@ -97,51 +108,43 @@ where
...
-            // use network sender to send command
-            self.network_send.send(request)?;
-            if let Err(e) = receiver.await? {
-                return Err(anyhow!(
-                    "The bitswap failed, please check server logs {:?}",
-                    e
-                ));
+        let new = self.check_blockstore(root_cid).await?;
+        let dag = self.store.dag_traversal(&root_cid)?;
+        if new {
```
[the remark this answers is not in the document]
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/ebb750854c81e7ec038c6e740b0cc8db0d4a3085

crates/ursa-rpc-service/src/api.rs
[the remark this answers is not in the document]
ozwaldorf: Corrected the size to be the full car instead of blocks alone
- For bitswap: implemented a method on store as suggested `car_size(cid)`
- For origin: gets the length of the body and returns that to sync_content (formerly check_blockstore), which will call provide_cid with the size

https://github.com/fleek-network/ursa/pull/278/commits/ebb750854c81e7ec038c6e740b0cc8db0d4a3085

crates/ursa-rpc-service/src/api.rs
ozwaldorf: addressed here: https://github.com/fleek-network/ursa/pull/278/commits/ebb750854c81e7ec038c6e740b0cc8db0d4a3085

crates/ursa-rpc-service/src/api.rs
ozwaldorf: Renamed `check_blockstore` to `sync_content` here:
https://github.com/fleek-network/ursa/pull/278/commits/ebb750854c81e7ec038c6e740b0cc8db0d4a3085

crates/ursa-rpc-service/src/api.rs
```diff
@@ -97,51 +108,29 @@ where
...
-                    "The bitswap failed, please check server logs {:?}",
-                    e
-                ));
-            }
-        }
-        self.store.blockstore().get(&cid)
+    async fn get(&self, cid: Cid) -> Result<Vec<u8>> {
+        self.sync_content(cid).await?;
+        let content =
+            self.store.blockstore().get(&cid)?.ok_or_else(|| {
```
b0xtch, line 124:
```suggestion
            self.store.blockstore().get(&cid)
```

crates/ursa-rpc-service/src/api.rs
```diff
@@ -283,6 +241,177 @@ where
...
+    /// Ensure a root cid is synced to the blockstore
+    async fn sync_content(&self, cid: Cid) -> Result<()> {
+        if !self.store.blockstore().has(&cid)? {
+            info!("Requesting block with the cid {cid:?}");
+            if let Err(e) = self.get_network(cid).await {
+                info!("Failed to get content from network: {}", e);
+                let size = self.get_origin(cid).await?;
+                self.provide_cid(cid, Some(size)).await
+            } else {
+                self.provide_cid(cid, None).await
```
b0xtch, line 273: isn't `size` required for the announcement?

crates/ursa-rpc-service/src/api.rs
```diff
@@ -283,6 +241,177 @@ where
...
+            }
+        } else {
+            Ok(())
+        }
```
b0xtch, line 277:
```suggestion
        if !self.store.blockstore().has(&cid)? {
            info!("Requesting block with the cid {cid:?}");
            if let Err(e) = self.get_network(cid).await {
                info!("Failed to get content from network: {}", e);
                let size = self.get_origin(cid).await?;
            } 
            self.provide_cid(cid, Some(size)).await
        }
        Ok(())
```

if the case above has to hold something like this where we have to enforce size

crates/ursa-rpc-service/src/api.rs
ozwaldorf: If size is None, then `provide_cid` will construct a temporary file and get the length using store.car_size(cid)

crates/ursa-store/src/store.rs
```diff
@@ -169,6 +178,32 @@ where
         }
         Ok(res)
     }
+
+    /// Build a temporary car file from a root cid and return the size
+    async fn car_size(&self, root_cid: &Cid) -> Result<u64> {
```
b0xtch, line 183: This behavior is expensive, have we benchmarked this?

crates/ursa-rpc-service/src/api.rs
b0xtch: I don't follow, bitswap relationship with announcement size

b0xtch: Size shouldn't be None

ozwaldorf: With origin, we get back a car response body, so we just .len() and done

With bitswap, fetched content is block by block and doesn't return a car size to use. So provide_cid accepts either a given size, or None which will make it calculate the car size on its own.

https://github.com/fleek-network/ursa/blob/841b87224aeeea04dd1e7a160475a48eca23b0e0/crates/ursa-rpc-service/src/api.rs#L380-L383

b0xtch: I am not sure you are following the comment I made, the size should never be None right. The function should just provide the cid with the size. How it gets is addressed in another comment

crates/ursa-store/src/store.rs
ozwaldorf: It could probably be improved by manually measuring a car file and creating a formula to directly calculate the size from the content lengths instead of constructing a full car

crates/ursa-rpc-service/src/api.rs
ozwaldorf: size wouldn't be defined if bitswap is successful

b0xtch: I am not sure i follow, size is always defined no?

crates/ursa-store/src/store.rs
b0xtch: I think we should traverse the blockstore and get the blob size not create a new file

ozwaldorf: Don't we need the car file size not blob size for gateway tho?

crates/ursa-rpc-service/src/api.rs
ozwaldorf: Bitswap doesn't return any information about the car size of the content it fetched

In the code you suggested, size only exists in the scope of if bitswap failed and get origin succeeded

b0xtch: Still need more information why would bitswap return car file size?

b0xtch: In the code I suggest that size should always be some not none

crates/ursa-rpc-service/src/api.rs
ozwaldorf: ah okay, I can move the call to car_size and make provide_cid always accept a size

crates/ursa-rpc-service/src/api.rs
ozwaldorf: yup, realized what you were suggesting, will make the change to the method and move the logic

ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/171e54140367661a9e63e432d178b3ed9b0f876a

crates/ursa-rpc-service/src/api.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/171e54140367661a9e63e432d178b3ed9b0f876a

crates/ursa-rpc-service/src/api.rs
ozwaldorf: If sync content is Ok(()), error being returned early, then blockstore.get should never be empty

b0xtch: "?" Should take care of that then. Not a big deal just comment

ozwaldorf: Ah yeah unfortunately the verbosity is required since get -> Result<Option>>, so have to translate option into an error to `?` it

crates/ursa-store/src/store.rs
```diff
@@ -129,11 +133,16 @@ where
     }
 }
 
+#[async_trait]
 pub trait Dag {
     /// traverse a dag and get full dag given a root cid
     fn dag_traversal(&self, root_cid: &Cid) -> Result<Vec<(Cid, Vec<u8>)>>;
+
+    /// Build a temporary car file from a root cid and return the size
+    async fn car_size(&self, root_cid: &Cid) -> Result<u64>;
```
kckeiks, line 142: This trait seemed to provide a generic dag interface but this method seems specific to car files. Maybe it would fit better as a method of `UrsaStore`.

ozwaldorf: Hmm, I had initially thought it fit there since it directly uses traverse_dag to build the file, I can move it to ursastore directly but just double checking

crates/ursa-store/src/store.rs
ozwaldorf: made some improvements to the method
benchmarked 1, 10, 100 MB car files at 1ms, 12ms, and 150ms respectively with the optimizations

crates/ursa-store/src/store.rs
kckeiks: And now that I'm thinking about it. The only implementer of `Dag` is `UrsaStore` so I don't think we need abstract `dag_traversal` and `car_size`. I recommend that we simplify this and move these methods in `UrsaStore`.

ozwaldorf: done. I actually put it on BlockstoreExt for a more generic addition https://github.com/fleek-network/ursa/pull/278/commits/04b7d8f83c37e68f7c29250dc04f4e4d7e53793a

kckeiks:
Looking good. I see some unresolved conversations though, are we good to move forward with this PR?

ozwaldorf:
To unblock this I'll do as suggested here: https://github.com/fleek-network/ursa/pull/278#discussion_r1071470631
But I'm still unconvinced that we should skip over making sure the root cid is present. I'll put together a little showcase of the potential issue separately and address it from there

crates/ursa-rpc-service/src/api.rs
ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/dc750a096aeee288a693c195e2bbd34bd205227c

crates/ursa-store/src/store.rs
```diff
@@ -84,6 +87,58 @@ pub trait BlockstoreExt: Blockstore {
 
         Ok(cids)
     }
+
+    /// traverse a dag and get full dag given a root cid
+    fn dag_traversal(&self, root_cid: &Cid) -> Result<Vec<(Cid, Vec<u8>)>> {
```
b0xtch, line 92: Moving this here is not appropriate. This a blockstore extension, aimed at 

> /// Extension methods for inserting and retrieving IPLD data with CIDs

b0xtch:
> To unblock this I'll do as suggested here: [#278 (comment)](https://github.com/fleek-network/ursa/pull/278#discussion_r1071470631) But I'm still unconvinced that we should skip over making sure the root cid is present. I'll put together a little showcase of the potential issue separately and address it from there

You can still check the root cid, like suggested after using the `load_car` file. To see if it returns the same root cid. Then we can wait or pr this https://github.com/filecoin-project/ref-fvm/issues/1522

b0xtch (approved):

crates/ursa-store/src/store.rs
ozwaldorf: Wouldn't traversing a dag be `retrieving IPLD data with CIDs` ? Was suggested to move here:  https://github.com/fleek-network/ursa/pull/278#discussion_r1081651261

ozwaldorf: I can move it directly into UrsaStore, but figured it'd be more generic and usable if on the BlockstoreExt instead, as all you need is an inner store for the methods

kckeiks: > Wouldn't traversing a dag be `retrieving IPLD data with CIDs` ? Was suggested to move here: [#278 (comment)](https://github.com/fleek-network/ursa/pull/278#discussion_r1081651261)

I suggested that we move these to `UrsaStore` not `BlockstoreExt`.

kckeiks: > I can move it directly into UrsaStore, but figured it'd be more generic and usable if on the BlockstoreExt instead, as all you need is a store for the methods

What helps me decide whether a method should go in a trait is "does it make sense for all users of trait X to have to implement this behaviour/method?".

ozwaldorf: https://github.com/fleek-network/ursa/pull/278/commits/582dec7c6012b72a74513978db247b00184c9bdb

b0xtch (approved):

PR #283, opened 2023-01-11
feat: gateway cache conditional

### LINKED ISSUE #210: feat: gateway cache conditional

kckeiks:
## Why

We don't want to cache if the content is too large.

Fixes #210 

## What

Stream directly from node to client.

## Demo
Set `cache_control_max_size = 27000 `

Input:
> -rw-r--r--   1 acadia  staff   27044 Dec 18 16:50 basic.car

Command:
> curl https://0.0.0.0:80/bafybeifyjj2bjhtxmp235vlfeeiy7sz6rzyx3lervfk3ap2nyn4rggqgei 

Output:
```
  2023-01-13T20:36:44.316160Z  INFO ursa_gateway::worker::cache::server: Content size is 27044...skipping cache
    at crates/ursa-gateway/src/worker/cache/server.rs:64
    in ursa_gateway::worker::cache::server::Cache missed
    in ursa_gateway::server::route::api::v1::get::Get car handler
    in axum_tracing_opentelemetry::middleware::trace_extractor::HTTP request with otel.name: GET /:cid, http.client_ip: , http.flavor: 2.0, http.host: , http.method: GET, http.route: /:cid, http.scheme: https, http.target: /bafybeifyjj2bjhtxmp235vlfeeiy7sz6rzyx3lervfk3ap2nyn4rggqgei, http.user_agent: curl/7.84.0, otel.kind: server, trace_id: 52b1d942d75ede5a77c7ef6d525f94ea


```

## Checklist

- N/A I have made corresponding changes to the tests

### DIFF
--- a/crates/ursa-gateway/src/resolver/mod.rs
+++ b/crates/ursa-gateway/src/resolver/mod.rs
@@ -24,7 +24,7 @@ pub struct Resolver {
 }
 
 #[derive(Debug)]
-pub struct Response {
+pub struct NodeResponse {
     pub resp: response::Response<Body>,
     pub size: u64,
 }
@@ -37,7 +37,7 @@ impl Resolver {
         }
     }
 
-    pub async fn resolve_content(&self, cid: &str) -> Result<Response, Error> {
+    pub async fn resolve_content(&self, cid: &str) -> Result<NodeResponse, Error> {
         let endpoint = format!("{}/{cid}", self.indexer_cid_url);
 
         let uri = endpoint.parse::<Uri>().map_err(|e| {
@@ -172,7 +172,7 @@ impl Resolver {
             };
             match self.client.get(uri).await {
                 Ok(resp) => {
-                    return Ok(Response {
+                    return Ok(NodeResponse {
                         resp,
                         size: metadata.size,
                     })
--- a/crates/ursa-gateway/src/resolver/mod.rs
+++ b/crates/ursa-gateway/src/resolver/mod.rs
@@ -1,7 +1,7 @@
 pub mod model;
 
 use anyhow::{anyhow, Context};
-use axum::{body::Body, http::response::Parts, response::Response};
+use axum::{body::Body, http::response, http::response::Parts};
 use hyper::{body::to_bytes, client::HttpConnector, StatusCode, Uri};
 use hyper_tls::HttpsConnector;
 use libp2p::multiaddr::Protocol;
@@ -21,6 +21,12 @@ pub struct Resolver {
     client: Client,
 }
 
+#[derive(Debug)]
+pub struct Response {
+    pub resp: response::Response<Body>,
+    pub size: u64,
+}
+
 impl Resolver {
     pub fn new(indexer_cid_url: String, client: Client) -> Self {
         Self {
@@ -29,7 +35,7 @@ impl Resolver {
         }
     }
 
-    pub async fn resolve_content(&self, cid: &str) -> Result<Response<Body>, Error> {
+    pub async fn resolve_content(&self, cid: &str) -> Result<Response, Error> {
         let endpoint = format!("{}/{cid}", self.indexer_cid_url);
 
         let uri = endpoint.parse::<Uri>().map_err(|e| {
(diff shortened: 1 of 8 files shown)


crates/ursa-gateway/src/resolver/mod.rs
```diff
@@ -23,6 +23,12 @@ pub struct Resolver {
     client: Client,
 }
 
+#[derive(Debug)]
+pub struct Response {
```
scott-dn, line 27: Maybe `CachNodeResponse` for clarity, as `Response` is a bit vague.

crates/ursa-gateway/src/worker/cache/server.rs
```diff
@@ -2,44 +2,41 @@ use std::sync::Arc;
...
-        &self,
-        k: &str,
-        no_cache: bool,
-    ) -> Result<StreamBody<ReaderStream<DuplexStream>>, Error> {
-        let (mut w, r) = duplex(self.stream_buf as usize);
+    async fn get_announce(&self, k: &str, no_cache: bool) -> Result<StreamBody, Error> {
         if no_cache {
             let span = info_span!("Cache invalidate");
-            fetch_and_insert(k, &self.tx, w).instrument(span).await?;
+            fetch(k, &self.tx)
```
scott-dn, line 34: Essentially, `no cache` or `cache miss` are the same, right?

crates/ursa-gateway/src/resolver/mod.rs
kckeiks: That's pretty verbose. How about `NodeResponse`? :)

crates/ursa-gateway/src/worker/cache/server.rs
kckeiks: I didn't noticed that the argument was coming from the HTTP cache control header. Will update.

crates/ursa-gateway/src/worker/cache/server.rs
```diff
@@ -82,14 +81,11 @@ async fn fetch_and_insert(
...
-            error!("Failed to receive response from resolver: {e:?}");
-            anyhow!("Failed to receive response from resolver")
-        })??
-        .into_parts()
-    {
+    let response = rx.await.map_err(|e| {
+        error!("Failed to receive response from resolver: {e:?}");
+        anyhow!("Failed to receive response from resolver")
+    })??;
+    let mut body = match response.resp.into_parts() {
```
b0xtch, line 88: This is an interesting behavior, getting the complete data first and then checking if it's larger than the specified limit for transmission, isn't it better to do a pre-check on the size of the requested cid. 

Do correct me if I am wrong on functionality. Maybe there are other better alternatives but a pre-check is the first one that come to mind.

b0xtch: if you already have the data, then why not stream it, think at scale too if the file is GBs worth. Also, this assumes a full content-length grab. instead of a byte range or a stream

b0xtch: Technically you will never cross the 1GB mark then if all files are chunked and streamed in that manner. Food for thought

kckeiks: True but we know the size of the content because it is advertised (#221) to the global indexer so we have that data point before we have all of the data. `Response.size` is a value returned by indexer.

b0xtch: which isn't a trusted value. this can be improved in a later PR, but to note.

scott-dn: > which isn't a trusted value. this can be improved in a later PR, but to note.

agree, we can create an issue for this and PR later.

scott-dn: created #291, to track this issue.

scott-dn (approved):
Lgtm

PR #286, opened 2023-01-12
chore: add node abstraction

kckeiks:
## Why

Passes node service state to other tests after bootstrapping.

## Checklist

- N/A I have made corresponding changes to the tests
- N/A I have made corresponding changes to the documentation
- N/A I have run the app using my feature and ensured that no functionality is broken

### DIFF
--- a/test-plans/data-transfer/Cargo.toml
+++ b/test-plans/data-transfer/Cargo.toml
@@ -4,8 +4,11 @@ name = "data-transfer"
 version = "0.1.0"
 
 [dependencies]
+cid = "0.8.6"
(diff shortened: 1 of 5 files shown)


matthias-wright:
Looks great, left a comment

kckeiks:
> Looks great, left a comment

Where is the comment?

matthias-wright:
Weird that it is not visible to you. The node struct has a graphsync store and I was wondering if we should also add a bitswap store.

matthias-wright (approved):
lgtm

Issue #271, opened 2023-01-09
test: add content discovery/fetching test plan


kckeiks:
# Description

When node does not have data in local storage, it retrieves using bloom filters and bitswap.

PR #322, opened 2023-01-23
test: content discovery/fetching test plan

matthias-wright:
## Why
See https://github.com/fleek-network/ursa/issues/271. This allows us to benchmark how fast nodes can retrieve content from their peers using bitswap.

## What
- Added new test plan to test and benchmark bitswap retrieval from peers.

## Checklist

- [N/A] I have made corresponding changes to the tests

### DIFF
--- a/test-plans/fetching/manifest.toml
+++ b/test-plans/fetching/manifest.toml
@@ -1,0 +1,15 @@
+name = "fetching"
+
+[defaults]
+builder = "docker:generic"
+runner = "local:docker"
+
+[builders."docker:generic"]
+enabled = true
+
+[runners."local:docker"]
+enabled = true
+
+[[testcases]]
+name = "fetching"
+instances = { min = 2, max = 30, default = 5 }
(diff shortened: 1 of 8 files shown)


test-plans/fetching/manifest.toml
```diff
@@ -0,0 +1,15 @@
...
+builder = "docker:generic"
+runner = "local:docker"
+
+[builders."docker:generic"]
+enabled = true
+
+[runners."local:docker"]
+enabled = true
+
+[[testcases]]
```
kckeiks, line 13: What is this table for? Can we control which tests run with this?

kckeiks: I was thinking last night if we might be able to access this value in the code, then we can control which tests run and we might not need to split tests. What do you think?

matthias-wright: If there are multiple binaries within one test plan, they are listed there.

matthias-wright: For each binary, we can specify different test cases and also specify parameters for those test cases.

kckeiks (approved):
lgtm

PR #334, opened 2023-01-24
chore: domain migration to `server_config` from `provider_config`

### LINKED ISSUE #244: chore: Move domain out of the ProviderConfig
After #233(unifying the servers), the domain is not specific to index provider and can be moved to server config

ozwaldorf:
## Why

First part of migrating domain from provider_config -> server_config. After some time the provider config field and migration logic can be removed completely

Fixes #244 

## What

- add server_config.domain
- make old provider_config.domain optional 
- when loading an existing file, and the field is populated, move it to new location

## Demo

 

image

## Notes

- After some time, `provider_config.domain` can be completely removed
- I set the default value for domain to be `ip4/127.0.0.1/tcp/4069` for new nodes, so as to have a concrete `Multiaddr` type for the domain field. The side effect of new nodes getting this value may not be desired and if not, I can set the type back to string and have it default as empty

## Checklist

### DIFF
--- a/crates/ursa-index-provider/src/config.rs
+++ b/crates/ursa-index-provider/src/config.rs
@@ -3,9 +3,10 @@ use std::path::PathBuf;
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ProviderConfig {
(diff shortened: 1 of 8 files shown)


matthias-wright (approved):
lgtm :rocket:

kckeiks (approved):

PR #336, opened 2023-01-25
chore: move kad replication factor to config

matthias-wright:
## Why
The replication factor was hardcoded in.

## What
- Moves the kad replication factor into `NetworkConfig`.

## Checklist

### DIFF
--- a/crates/ursa-network/src/config.rs
+++ b/crates/ursa-network/src/config.rs
@@ -40,6 +40,9 @@ pub struct NetworkConfig {
     /// Defaults to devnet tracker.
     #[serde(default = "NetworkConfig::default_tracker")]
     pub tracker: String,
+    /// Determines the number of closest peers to which a record is replicated
+    #[serde(default = "NetworkConfig::default_kad_replication_factor")]
+    pub kad_replication_factor: usize,
 }
 
 impl NetworkConfig {
@@ -82,6 +85,9 @@ impl NetworkConfig {
     fn default_identity() -> String {
         "default".to_string()
     }
+    fn default_kad_replication_factor() -> usize {
+        8
+    }
 }
 
 impl Default for NetworkConfig {
@@ -98,6 +104,7 @@ impl Default for NetworkConfig {
             identity: Self::default_identity(),
             tracker: Self::default_tracker(),
             keystore_path: Self::default_keystore_path(),
+            kad_replication_factor: Self::default_kad_replication_factor(),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


ozwaldorf:
We're not using kademlia at all for key value store, or replication, right?

matthias-wright:
No, only discovery afaik.

ozwaldorf:
I'm not sure it even matters what's set there in that case, as kad would never actually do replication. If our actual replication behavior is configurable like this, we should provide that in the configs instead

matthias-wright:
Atm the actual replication behaviour is sending a caching request to all connected peers. Okay makes sense. There was a todo comment for this in the code though.

ozwaldorf:
We can probably just remove the comment, cc @b0xtch in case there was a specific reason for it

b0xtch (approved):

crates/ursa-network/src/config.rs
```diff
@@ -82,6 +85,9 @@ impl NetworkConfig {
     fn default_identity() -> String {
         "default".to_string()
     }
+    fn default_kad_replication_factor() -> usize {
+        8
```
mxinden, line 89: May I ask why `8` is chosen here? In other words, why you are deviating from the default?

For those interested, the following measurement report on the liveness of Provider Records in the IPFS DHT might be helpful https://github.com/protocol/network-measurements/blob/master/results/rfm17-provider-record-liveness.md.

mxinden (commented):
Unsolicited review. Hope you don't mind.

I guess the question below doesn't quite matter given that, as you said, you don't use neither the Record nor the Provider Record system.

crates/ursa-network/src/config.rs
matthias-wright: Hi @mxinden, thanks for reaching out, your input is always welcome here :)
We only use kademlia for peer discovery right now, but we are still considering to use it for content routing at some point. The default is set to 8 because that is the desired outbound degree for the meshes. Thanks for sharing the report, I'll definitely have a look!

mxinden: > The default is set to 8 because that is the desired outbound degree for the meshes.

Can you expand on "desired outbound degree for the meshes"? What is an "outbound degree" here? What are you referring to with "meshes"? Sorry in case I am missing something obvious.

The replication factor should (among other things) depend on the expected peer churn. If I publish a value today to X peers, will one of those X peers still be online when someone else queries for that value later on.

PR #340, opened 2023-01-25
feat(tracker): only provide nodes that registered within 30 days

ozwaldorf:
## Why

Tracker is returning a ton of stale nodes to prometheus. As a first effort to curate the list of "active" nodes, we should limit to 30 days, so that old nodes/configs eventually don't get scraped

This cuts the scrape target list down from ~1500 targets to ~1100 targets, and I'm sure many more being excluded as time progresses

## Notes

- We could probably use a shorter time period, like 2-3 weeks, but I went with a more generous number to avoid excluding nodes that just don't update/restart often and wouldn't reregister in time

## Checklist

- [n.a] I have made corresponding changes to the tests
- [n.a] I have made corresponding changes to the documentation

### DIFF
--- a/crates/ursa-tracker/src/main.rs
+++ b/crates/ursa-tracker/src/main.rs
@@ -8,6 +8,7 @@ use hyper::HeaderMap;
 use rocksdb::{IteratorMode, WriteBatch, DB};
 use serde_json::{json, Value};
 use std::{env, net::SocketAddr, sync::Arc};
+use std::time::{SystemTime, UNIX_EPOCH};
 use tracing::{error, info};
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter};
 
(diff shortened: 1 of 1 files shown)


crates/ursa-tracker/src/main.rs
```diff
@@ -105,12 +110,17 @@ async fn registration_handler(
...
+        .unwrap()
+        .as_millis() as u64;
     let services: Vec<PrometheusDiscoveryChunk> = db
         .iterator(IteratorMode::Start)
         .filter_map(|res| {
             if let Ok((_, v)) = res {
                 let node: Node = serde_json::from_slice(&v).unwrap();
-                if node.telemetry {
+                // only return registrations in the last 1 month
+                if now - node.last_registered < 2629800000 {
```
matthias-wright, line 123: Should we still check for `node.telemetry`, so `node.telemetry && now - node.last_registered < 2629800000` or is that not necessary?

ozwaldorf: Don't think we really need to check `node.telemetry`, it's always set to true. It was from before when metrics and rpc were separated, but now their unified so a node will always expose their metrics

matthias-wright (approved):

PR #341, opened 2023-01-26
feat(proxy): add caches

kckeiks:
## Checklist

### DIFF
--- a/crates/ursa-proxy/src/cache/tlrfu/mod.rs
+++ b/crates/ursa-proxy/src/cache/tlrfu/mod.rs
@@ -29,7 +29,6 @@ pub trait ByteSize {
     fn len(&self) -> usize;
 }
 
-#[cfg(not(test))]
 pub fn now() -> SystemTime {
     SystemTime::now()
 }
--- a/crates/ursa-proxy/src/cache/tlrfu/mod.rs
+++ b/crates/ursa-proxy/src/cache/tlrfu/mod.rs
@@ -1,0 +1,192 @@
+mod lru;
+
+use std::{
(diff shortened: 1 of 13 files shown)


crates/ursa-proxy/src/main.rs
```diff
@@ -1,22 +1,27 @@
...
-mod core;
-
 #[tokio::main]
 async fn main() -> Result<()> {
     let Cli {
         command: Commands::Daemon(opts),
     } = Cli::parse();
     let config = load_config(&opts.config.parse::<PathBuf>()?)?;
-    ProxyCore::new(config).start_servers().await.unwrap();
+    let cache = TlrfuCache::new(200_000_000, 5 * 60 * 1000, 2_000_000);
```
b0xtch, line 21: Giving these numbers types/parametrize them will help readability.

crates/ursa-proxy/src/cache/tlrfu/mod.rs
```diff
@@ -0,0 +1,185 @@
+mod lru;
```
kckeiks, line 1: Copypasta from gateway. Maybe this should be released as a crate so we can just wrap it up like moka. Thoughts?

b0xtch: That's sounds like a good idea, I see this can be used for other purposes. I would cc @scott-dn 

Maybe not in this PR

kckeiks: Sounds good

kckeiks:
Closing in favor of https://github.com/fleek-network/ursa/pull/345.

PR #343, opened 2023-01-26
feat: bootstrap timeout retry logic + new peer log level

ozwaldorf:
## Why

Some node operators have been reporting random bootstrap timeouts, and metrics show some nodes with low/no peers.

## What

- adds retry logic for bootstrap timeouts
- gives up after 10 tries
- raise new peer log level for more comfort to node operators that their node is receiving peers correctly

## Notes

I raised the log level for new peers, but it may be preferable to revert that change. If so, we could start directing node runners to use
```
curl localhost:4069/metrics | grep peers 
```
which would return something like 
```
# TYPE swarm_connected_peers gauge
swarm_connected_peers 101
```
as a small healthcheck on the nodes connection status

## Checklist

### DIFF
--- a/crates/ursa-network/src/behaviour.rs
+++ b/crates/ursa-network/src/behaviour.rs
@@ -18,6 +18,7 @@ use db::Store;
 use fvm_ipld_blockstore::Blockstore;
 use graphsync::GraphSync;
 use libipld::{store::StoreParams, Cid};
+use libp2p::kad::{NoKnownPeers, QueryId};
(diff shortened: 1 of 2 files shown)


mxinden:
If I understand correctly, you are running `Kademlia::bootstrap` once (or multiple times on failure). In my eyes the name of the method is misleading. In other words, it is worth considering running `Kademlia::bootstrap` on an interval across the entire lifecycle of a node. That should also mirror the behavior of Kubo. If I recall correctly they call it "random walk".

PR #344, opened 2023-01-27
feat: random kademlia walks

### LINKED ISSUE #73: chore: automatic kademlia walks

ozwaldorf:
## Why

Resolves #73 

## What

- add `network_config.kad_walk_interval` (defaulting to 300 seconds / 5 minutes)
- add timer to trigger random kad walks in service.rs

## Demo

(for the demo, lowered the delay to 1s just to grab a screenshot quickly)

image

## Checklist

- ~~I have made corresponding changes to the tests~~
- ~~I have made corresponding changes to the documentation~~

### DIFF
--- a/crates/ursa-network/src/service.rs
+++ b/crates/ursa-network/src/service.rs
@@ -220,7 +220,7 @@ where
     cached_content: CacheSummary,
     /// Content summaries from other nodes.
     peer_cached_content: HashMap<PeerId, CacheSummary>,
-    /// Interval for random Kademlia walks
+    /// Interval for random Kademlia walks.
     kad_walk_interval: u64,
 }
 
--- a/crates/ursa-network/src/service.rs
+++ b/crates/ursa-network/src/service.rs
@@ -40,11 +40,12 @@ use libp2p::{
 };
 use libp2p_bitswap::{BitswapEvent, QueryId};
 use rand::prelude::SliceRandom;
-use std::fmt::Debug;
 use std::{
     collections::{HashMap, HashSet},
+    fmt::Debug,
     num::{NonZeroU8, NonZeroUsize},
     sync::Arc,
+    time::Duration,
 };
 use tokio::{
     select,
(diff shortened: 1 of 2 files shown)


ozwaldorf:
cc @mxinden continuing the conversation from https://github.com/fleek-network/ursa/pull/343#issuecomment-1406789107

https://github.com/fleek-network/ursa/blob/c4f3d832dc7737dceec5d3e8de227b98b329dab9/crates/ursa-network/src/service.rs#L949-L951
In this pr I used get closest peers to a random peer id, whereas you previously suggested to use bootstrap() at an interval. Depending on the implementation, I've seen both kinds used, I'm curious if there's any real difference or benefit from using bootstrap directly for the walk

crates/ursa-network/src/service.rs
```diff
@@ -218,6 +220,8 @@ where
     cached_content: CacheSummary,
     /// Content summaries from other nodes.
     peer_cached_content: HashMap<PeerId, CacheSummary>,
+    /// Interval for random Kademlia walks
```
kckeiks, line 223:
```suggestion
    /// Interval for random Kademlia walks.
```

kckeiks (approved):

crates/ursa-network/src/service.rs
kckeiks: Don't worry about this one. I'm planning to do some clean up with the comments soon.

kckeiks (approved):

mxinden:
Change looks good to me.

> @mxinden continuing the conversation from [#343 (comment)](https://github.com/fleek-network/ursa/pull/343#issuecomment-1406789107). you suggested there to use bootstrap() at an interval, whereas in this pr I used get closest peers to a random peer id as seen here:
> 
> https://github.com/fleek-network/ursa/blob/c4f3d832dc7737dceec5d3e8de227b98b329dab9/crates/ursa-network/src/service.rs#L949-L951
> 
> Checking out several other implementations, I've seen both flows used interchangeably to keep the routing table fresh. I'm curious if there's any real difference or benefit from using bootstrap directly for the walk?
> 
> And of course, as always really appreciate the reviews and feedback! pray

A random walk does a single DHT lookup. A `bootstrap` call does one lookup per bucket.

Unfortunately I don't have any data on which of the two should happen periodically. Ultimately you are trading resource usage (periodic lookups) for low-latency (due to up-to-date routing table). Until we have more data, I suggest doing the cheaper operation, i.e. a random walk as you do in this pull request.

PR #363, opened 2023-02-06
feat: add testground to ci [work in progress]

matthias-wright:
See #311.
This is a draft PR because it's not working currently. I'm running into https://github.com/testground/testground/issues/1541. Interestingly, this only happens when using Github workflows and never when running the test plans locally. The problem is that the directories for the current run are not created at `/root/.config/testground/data/outputs/local_docker/fetching/<run-id>/<plan>/<index>`. The directory `/root/.config/testground/data/outputs` exists. Manually creating these directories might be a quick fix but won't solve the underlying issue.
I created a discussion in the testground repo for this issue: https://github.com/testground/testground/discussions/1568

Another thing to note:
At the moment, we are pulling the ursa crates from the ursa git repository. This can be seen for example here: https://github.com/fleek-network/ursa/blob/main/test-plans/fetching/Cargo.toml. This is fine but defeats the purpose of running the test plans on every PR. For this, we have to use the code from the current commit. The most straightforward way that I came up with to achieve this is to copy the code from the ursa crates into the test plan directories and then temporarily overwriting the corresponding Cargo.toml.

I added a script `scripts/run_testground.sh` that takes care of this when running testground locally.

### DIFF
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -1,0 +1,73 @@
+name: testground-ci
+
+on:
+  push:
+    branches: [ "main" ]
+    paths:
+      - "crates/**"
+      - "test-plans/**"
+      - "Cargo.toml"
+      - ".github/workflows/testground.yml"
+  pull_request:
+    branches: [ "main" ]
+    paths:
+      - "crates/**"
+      - "test-plans/**"
+      - "Cargo.toml"
+      - ".github/workflows/testground.yml"
+
+jobs:
+  tests:
+    name: test plan ${{ matrix.plan }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        plan: ["data-transfer", "fetching"]
+
+    steps:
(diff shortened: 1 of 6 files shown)


kckeiks:
Thanks for taking this on!

Will the CI only validate that the tests build or that they ran successfully? The latter will be more complex right?

matthias-wright:
Thanks for taking a look! If the test plans would actually run without errors, then the CI will validate that they run successfully. In [this](https://github.com/fleek-network/ursa/blob/43862f8806f87f98430eefcf355f7b12b1391bbc/.github/workflows/testground.yml#L70) line, the `--wait` argument will make the command wait for the test plan to complete and the exit status will be 0 only if the test plan succeeded. The github ci then picks up on the exit status, so it should all work out.

b0xtch:
looks good

jxs:
Hey folks, this might interest you: https://github.com/libp2p/test-plans/issues/103

b0xtch:
Should we close this, given the move away from test ground?

PR #372, opened 2023-02-07
refactor(store): remove GraphSyncStorage

ozwaldorf:
## Why

The explicit `GraphSyncStore` struct is not needed, as the required trait for graphsync can be instead implemented directly on UrsaStore itself 

## What

- remove graphsyncstorage and use ursastore directly
- construct bitswapStorage from behaviour from a single ursastore param

## Checklist

### DIFF
--- a/crates/ursa-store/src/store.rs
+++ b/crates/ursa-store/src/store.rs
@@ -20,6 +20,14 @@ pub struct UrsaStore<S> {
     pub db: Arc<S>,
 }
 
+impl<S> Clone for UrsaStore<S> {
+    fn clone(&self) -> Self {
+        Self {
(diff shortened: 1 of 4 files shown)


crates/ursa-store/src/store.rs
```diff
@@ -20,6 +20,14 @@ pub struct UrsaStore<S> {
     pub db: Arc<S>,
 }
 
+impl<S> Clone for UrsaStore<S> {
```
kckeiks, line 23: How about `#[derive(Clone)]` instead?

kckeiks:
Small nit otherwise looks good!

kckeiks (approved):

PR #373, opened 2023-02-08
chore(proxy): clean up

kckeiks:
## Why

I initially wanted to implement an event driven pattern where proxy would send events to cache using a channel. However, this slowed down the proxy (not by much though, but still, we don't want that). Therefore, I gave up on the event driven pattern and I extended the cache trait. I did some other clean up too and removed unnecessary code. Overall performance is still the same.

## What

- Update cache trait.
- Add more configuration options for moka.
- Moka handles its own ttl so I got rid of the timer worker.
- Fixed a bug that would panic when we had invalid data in config instead of safe shutting down.

## Demo

 
Running on release mode.

```
Summary:
  Success rate: 1.0000
  Total:        1.5492 secs
  Slowest:      0.0268 secs
  Fastest:      0.0001 secs
  Average:      0.0008 secs
  Requests/sec: 64547.3651

  Total data:   2.52 GiB
  Size/request: 26.41 KiB
  Size/sec:     1.63 GiB

Response time histogram:
  0.000 [1]     |
  0.003 [99702] |■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.005 [181]   |
  0.008 [38]    |
  0.011 [15]    |
  0.013 [12]    |
  0.016 [1]     |
  0.019 [0]     |
  0.021 [1]     |
  0.024 [41]    |
  0.027 [8]     |

Latency distribution:
  10% in 0.0004 secs
  25% in 0.0005 secs
  50% in 0.0007 secs
  75% in 0.0009 secs
  90% in 0.0011 secs
  95% in 0.0013 secs
  99% in 0.0018 secs

Details (average, fastest, slowest):
  DNS+dialup:   0.0216 secs, 0.0193 secs, 0.0251 secs
  DNS-lookup:   0.0000 secs, 0.0000 secs, 0.0001 secs

Status code distribution:
  [200] 100000 responses
```
## Checklist


crates/ursa-proxy/src/cache/moka_cache.rs
```diff
@@ -18,42 +22,40 @@ pub struct MokaCache {
...
-                        if let Err(e) = w.write_all(data.as_ref()).await {
-                            warn!("Failed to write to stream: {e:?}");
-                        }
-                    });
-                    response = Some(StreamBody::new(ReaderStream::new(r)).into_response());
+    fn get(&self, key: String) -> Option<Response> {
+        let mut response = None;
+        if let Some(data) = self.inner.get(&key) {
+            let (mut w, r) = duplex(self.stream_buf as usize);
+            spawn(async move {
```
b0xtch, line 43: nit: `task::spawn` consistency

b0xtch (approved):
Mint

crates/ursa-proxy/src/config.rs
```diff
@@ -22,21 +22,26 @@ pub struct ProxyConfig {
...
+    pub time_to_idle: u64,
+    pub time_to_live: u64,
 }
 
 impl Default for MokaConfig {
     fn default() -> Self {
         Self {
-            stream_buf: 2_000_000, // 2MB
+            max_capacity: 200_000_000,   //  Number of entries.
+            stream_buf: 1_000_000_000,   // 1GB.
```
ozwaldorf, line 42: Is `stream_buf` the proxies chunking size ?

kckeiks: Chunking size? 

This is what it's for https://docs.rs/tokio/latest/tokio/io/fn.duplex.html
> The max_buf_size argument is the maximum amount of bytes that can be written to a side before the write returns Poll::Pending.

ozwaldorf: I was thinking it was the chunk size used when streaming data to the client. No worries, seems like it's just an arbitrary number

ozwaldorf (approved):

scott-dn (approved):
Lgtm

ozwaldorf (approved):

scott-dn (approved):

PR #374, opened 2023-02-08
refactor: use git commit hash for agent versions

ozwaldorf:
## Why

Currently every node is identifying with the same crate version, which isn't updated, released, etc. We should instead use the git commit hash to identify the version of the codebase a node is running

## What

- nodes will use `ursa/<hash>` for their agent and package versions
- tests will use `ursa/tests` for their agent to provide some sort of separation for tests that connect to actual nodes
- Global metrics will now provide us real insights into what versions nodes are running

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ pem = "1.1.0"
 prometheus-client = "0.19.0"
 rand = "0.8.5"
 resolve-path = "0.1.0"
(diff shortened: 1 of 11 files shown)


kckeiks (dismissed):

scott-dn (dismissed):

ozwaldorf:
cc @b0xtch

PR #380, opened 2023-02-09
chore: remove port from domain

b0xtch:
## Why
- nodes expose 443 and not 4069

## What

- Change domain default
- We are not golang so using full words
- change gateway default indexer URL

## Checklist

### DIFF
--- a/crates/ursa-index-provider/src/provider.rs
+++ b/crates/ursa-index-provider/src/provider.rs
@@ -165,7 +165,7 @@ where
         if let Some(head_cid) = *self.head.read().unwrap() {
             let message = Message {
                 Cid: head_cid,
-                Addrs: vec![multiaddr],
+                Addresses: vec![multiaddr],
                 ExtraData: *b"",
             };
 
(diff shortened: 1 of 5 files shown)


crates/ursa-index-provider/src/provider.rs
```diff
@@ -181,9 +181,10 @@ where
 #[derive(Serialize, Deserialize, Debug)]
 pub struct Message {
     pub Cid: Cid,
-    pub Addrs: Vec<Multiaddr>,
+    pub Addresses: Vec<Multiaddr>,
```
scott-dn, line 184: Is the announcement message struct changed from the indexer?
> https://docs.cid.contact/index-provider/overview
```go
{
	Cid:      cid.Cid
	Addrs     [][]byte
	ExtraData []byte `json:",omitempty"`
}
```

b0xtch: You are right, reverting, it irks me

PR #381, opened 2023-02-10
feat: save public address instead of requesting from autonat each time

ozwaldorf:
## Why

Allows for us to use the last known public address. Sometimes, the NAT status will fall back to private due to network idle, and the address is unknown

## What

- increase log level for private nat status, and expand the error message to hint that the user's setup is wrong and should be fixed
- save last received public address

## Checklist

- [na] I have made corresponding changes to the tests
- [na] I have made corresponding changes to the documentation

### DIFF
--- a/crates/ursa-network/src/behaviour.rs
+++ b/crates/ursa-network/src/behaviour.rs
@@ -260,10 +260,6 @@ where
         self.gossipsub.publish(topic, data)
     }
 
-    pub fn public_address(&self) -> Option<&Multiaddr> {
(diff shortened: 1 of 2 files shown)


b0xtch:
4069 shouldn’t be exposed publicly

ozwaldorf:
> 4069 shouldn’t be exposed publicly

Wdym? It isn't here

theBeardA (approved):

b0xtch (approved):

PR #382, opened 2023-02-10
chore: manually provide ip addresses to index provider

b0xtch:
## Why

Autonat does not deterministically get the IP address of the node. 

## What

- Change domain to use public IP address

## Checklist

### DIFF
--- a/crates/ursa-rpc-service/src/config.rs
+++ b/crates/ursa-rpc-service/src/config.rs
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 pub struct ServerConfig {
     /// Public IP address of the node, eg. `/ip4/127.0.0.1`
     #[serde(default = "ServerConfig::default_addresses")]
-    pub addresss: Vec<Multiaddr>,
+    pub addresses: Vec<Multiaddr>,
     /// Port to listen on
     #[serde(default = "ServerConfig::default_port")]
     pub port: u16,
@@ -31,7 +31,7 @@ impl ServerConfig {
 impl Default for ServerConfig {
     fn default() -> Self {
         Self {
-            addresss: Self::default_addresses(),
+            addresses: Self::default_addresses(),
             port: Self::default_port(),
             addr: Self::default_addr(),
             origin: Default::default(),
--- a/crates/ursa-rpc-service/src/config.rs
+++ b/crates/ursa-rpc-service/src/config.rs
@@ -3,9 +3,9 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Deserialize, Serialize, Debug)]
 pub struct ServerConfig {
-    /// Domain Multiaddress of the node, eg. `/dns/test-node.ursa.earth`
-    #[serde(default = "ServerConfig::default_domain")]
-    pub domain: Multiaddr,
+    /// Public IP address of the node, eg. `/ip4/127.0.0.1`
+    #[serde(default = "ServerConfig::default_addresses")]
+    pub addresss: Vec<Multiaddr>,
     /// Port to listen on
     #[serde(default = "ServerConfig::default_port")]
     pub port: u16,
@@ -17,8 +17,8 @@ pub struct ServerConfig {
 }
 
 impl ServerConfig {
-    fn default_domain() -> Multiaddr {
-        "/ip4/127.0.0.1/tcp/4069".parse().unwrap()
+    fn default_addresses() -> Vec<Multiaddr> {
+        vec!["/ip4/127.0.0.1".parse().unwrap()]
     }
     fn default_port() -> u16 {
         4069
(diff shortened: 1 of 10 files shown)


ozwaldorf:
we should migrate old configs to new, since domain is just a multiaddr, we can move it to address as a single vec. Also note that domain was migrated out of provider config and into server config

b0xtch:
> we should migrate old configs to new, since domain is just a multiaddr, we can easily move it to address as a single entry vec. Also note that domain was migrated out of provider config and into server config

can you elaborate, what old vs new?

ozwaldorf:
> can you elaborate, what old vs new?

For nodes properly setup with a domain, they'd already have something configured like
```
[server_config]
domain = "/dns/foo.bar/" # recently migrated out of provider_config since its specific to the entire node
...
```

Since the new `provider_config.addresses` is just a vector of addresses to use for the announcement, we could reuse the domain value configured already right?

b0xtch:
> > can you elaborate, what old vs new?
> 
> For nodes properly setup with a domain, they'd already have something configured like
> 
> ```
> [server_config]
> domain = "/dns/foo.bar/" # recently migrated out of provider_config since its specific to the entire node
> ...
> ```
> 
> Since the new `provider_config.addresses` is just a vector of addresses to use for the announcement, we could reuse the domain value configured already right?

they would need to move to using IP addresses

crates/ursa/src/config.rs
```diff
@@ -34,15 +34,7 @@ impl UrsaConfig {
...
 
-            let mut config: UrsaConfig =
-                toml::from_str(&raw).context("Failed to parse config file")?;
-
-            // TEMP: remove this after some time
-            if let Some(domain) = config.provider_config.domain.clone() {
-                warn!("Config `provider_config.domain` depreciated, moving value to `server_config.domain`");
-                config.server_config.domain = domain.parse().unwrap();
-                config.provider_config.domain = None;
-            }
```
ozwaldorf, line 45: Also double checking, do you think 2 weeks is sufficient time to properly migrate this for node runners?
https://github.com/fleek-network/ursa/pull/334

b0xtch: yes

crates/ursa-rpc-service/src/tests/mod.rs
```diff
@@ -44,16 +43,14 @@ pub fn init() -> InitResult {
...
 
     let provider_engine = ProviderEngine::new(
         keypair,
         Arc::clone(&store),
         get_store(),
         ProviderConfig::default(),
         service.command_sender(),
-        server_address,
-        "/ip4/127.0.0.1/tcp/4069".parse().unwrap(),
+        vec!["/ip4/127.0.0.1/tcp/4069".parse().unwrap()],
```
ozwaldorf, line 53: For ease of maintenance this might be better
```suggestion
        ServerConfig::default_addresses(),
```

crates/ursa-rpc-service/src/config.rs
```diff
@@ -3,9 +3,9 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Deserialize, Serialize, Debug)]
 pub struct ServerConfig {
-    /// Domain Multiaddress of the node, eg. `/dns/test-node.ursa.earth`
-    #[serde(default = "ServerConfig::default_domain")]
-    pub domain: Multiaddr,
+    /// Public IP address of the node, eg. `/ip4/127.0.0.1`
+    #[serde(default = "ServerConfig::default_addresses")]
+    pub addresss: Vec<Multiaddr>,
```
ozwaldorf, line 8:
```suggestion
    pub addresses: Vec<Multiaddr>,
```

crates/ursa-rpc-service/src/config.rs
```diff
@@ -31,7 +31,7 @@ impl ServerConfig {
 impl Default for ServerConfig {
     fn default() -> Self {
         Self {
-            domain: Self::default_domain(),
+            addresss: Self::default_addresses(),
```
ozwaldorf, line 34:
```suggestion
            addresses: Self::default_addresses(),
```

crates/ursa/src/main.rs
```diff
@@ -94,21 +93,14 @@ async fn main() -> Result<()> {
...
                 let index_store = Arc::new(UrsaStore::new(Arc::clone(&Arc::new(provider_db))));
                 let index_provider_engine = ProviderEngine::new(
                     keypair,
                     Arc::clone(&store),
                     index_store,
                     provider_config,
                     service.command_sender(),
-                    server_address,
-                    server_config.domain.clone(),
+                    server_config.addresss.clone(),
```
ozwaldorf, line 103:
```suggestion
                    server_config.addresses.clone(),
```

theBeardA (approved):

kckeiks:
Looks good. I was talking with @theBeardA about which port number the gateway should use. If we plan to always encrypt the connection between gateway and node (I think, correct me if im wrong please) using HTTPS, 443 is a good default. If the port will not be advertised anymore, we should update the gateway here https://github.com/fleek-network/ursa/blob/main/crates/ursa-gateway/src/resolver/mod.rs#L124.

Issue #357, opened 2023-02-01
chore: large number of CacheRequests being sent


kckeiks:
# Description

There are currently too many CacheRequest being sent by a node.

# Checklist

- [x] ~I have ensured that my version is up-to-date~
- [x] ~I have ensured that my issue is reproducible~
- [x] I have ensured that my issue is not a duplicate

PR #383, opened 2023-02-11
feat(proxy): TLS config reload

### LINKED ISSUE #330: feat(proxy): add let's encrypt compatibility

kckeiks:
## Why

Fixes #330 

## What

Reload TLS config via the admin.

## Checklist


kckeiks:
Now that #357 and #127 are done. I'll get back to working on this.

crates/ursa-proxy/src/config.rs
```diff
@@ -58,3 +65,9 @@ impl Default for AdminConfig {
         }
     }
 }
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct TlsConfig {
+    pub cert_path: String,
+    pub key_path: String,
```
b0xtch, line 72: could we perhaps use `PathBuf`

crates/ursa-proxy/src/core/handler.rs
```diff
@@ -83,3 +172,20 @@ pub async fn proxy_pass<C: Cache>(
     };
     StreamBody::new(ReaderStream::new(reader)).into_response()
 }
+
+async fn serve_file_handler(
```
b0xtch, line 176: are we serving car files?

b0xtch (dismissed):

crates/ursa-proxy/src/core/handler.rs
kckeiks: No, this is for the let's encrypt challenge, `certbot` needs to be able to serve files. It can be configured like in nginx.

crates/ursa-proxy/src/core/handler.rs
```diff
@@ -22,6 +27,90 @@ use tracing::{error, info, warn};
...
 
+pub async fn init_server_app<C: Cache>(
+    server_config: ServerConfig,
+    cache: C,
+    client: Client,
+) -> Router {
+    let mut user_apps = Router::new();
+    for location in server_config.clone().location {
+        let route_path = format!(
+            "/{}/*path",
```
ozwaldorf, line 38: could we match for `/.well-known/acme-challenge/` ? certbot will only request from there

crates/ursa-proxy/src/config.rs
```diff
@@ -19,12 +19,19 @@ pub struct ProxyConfig {
...
 
-#[derive(Deserialize, Serialize, Debug)]
+#[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct ServerConfig {
     pub proxy_pass: String,
     pub listen_addr: String,
-    pub cert_path: String,
-    pub key_path: String,
+    pub server_name: String,
+    pub location: Vec<Location>,
```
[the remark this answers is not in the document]
kckeiks: Default for a location is no location.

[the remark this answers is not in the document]
kckeiks: No it can be set up like it is now.

qti3e (changes requested):
Good work overall, just one high sev exploit that needs to be fixed before merging.

crates/ursa-proxy/src/core/handler.rs
```diff
@@ -83,3 +172,20 @@ pub async fn proxy_pass<C: Cache>(
...
 }
+
+async fn serve_file_handler(
+    OriginalUri(uri): OriginalUri,
+    Extension(root): Extension<String>,
+) -> Response {
+    let mut file_path = PathBuf::from_str(root.as_str()).unwrap();
+    file_path.push(uri.path().trim_start_matches('/'));
+    let path_str = file_path.as_os_str().to_str().unwrap();
+    println!("{path_str}");
```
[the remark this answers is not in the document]
kckeiks: Yeah I need to push after CI is fixed. Cleaned up this part.

crates/ursa-proxy/src/core/handler.rs
```diff
@@ -83,3 +172,20 @@ pub async fn proxy_pass<C: Cache>(
...
     StreamBody::new(ReaderStream::new(reader)).into_response()
 }
+
+async fn serve_file_handler(
+    OriginalUri(uri): OriginalUri,
+    Extension(root): Extension<String>,
+) -> Response {
+    let mut file_path = PathBuf::from_str(root.as_str()).unwrap();
+    file_path.push(uri.path().trim_start_matches('/'));
+    let path_str = file_path.as_os_str().to_str().unwrap();
```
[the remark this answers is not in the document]
kckeiks: Yes, this is possible. Can this happen with the current set up or does NGINX have some measures in place?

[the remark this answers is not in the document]
qti3e: I think it would be better to:
1. resolve the path using `use resolve_path::PathResolveExt`
2. double check if it does not contain .. after resolve.
3. double check if the final path to read from starts with the directory path from config using [`starts_with`](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.starts_with)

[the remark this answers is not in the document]
qti3e: @kckeiks nvm my previous comment, can we use `tower_http:: services::ServeDir` and eliminate this function? 

https://github.com/tokio-rs/axum/blob/main/examples/static-file-server/src/main.rs#L44

it already has protection in place
https://docs.rs/tower-http/latest/src/tower_http/services/fs/serve_dir/mod.rs.html#36

crates/ursa-proxy/src/core/handler.rs
b0xtch: nodes should only be serving content addressable

kckeiks: This is the same as what NGINX is currently doing for the `certbot`. No other file is being served.

kckeiks: I'll talk to Helder to see if we can improve the process of setting up the certs and if we can to avoid having nginx and the proxy serve any files. There is a thread on the networking channel from yesterday.

kckeiks:
Closing in favor of https://github.com/fleek-network/ursa/pull/455.

PR #389, opened 2023-02-13
feat(gateway): cache indexer response

### LINKED ISSUE #387: feat(gateway): cache indexer response
Cache responses from indexer.

## Notes
Gateways spend time finding providers and must create an index of providers. 

Building index will be handled in #174

kckeiks:
## Why

Fixes #387

## What

- Add cache in resolver and cache responses from indexer.
- Allow users to configure the cache.

## Checklist

- [x] ~I have made corresponding changes to the tests~
- [x] ~I have made corresponding changes to the documentation~

### DIFF
--- a/crates/ursa-gateway/Cargo.toml
+++ b/crates/ursa-gateway/Cargo.toml
@@ -30,3 +30,4 @@ libp2p.workspace = true
 ursa-telemetry = { path = "../ursa-telemetry" }
 tracing-opentelemetry.workspace = true
 axum-prometheus.workspace = true
+moka = "0.9"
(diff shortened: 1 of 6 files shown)


crates/ursa-gateway/Cargo.toml
```diff
@@ -30,3 +30,4 @@ libp2p.workspace = true
 ursa-telemetry = { path = "../ursa-telemetry" }
 tracing-opentelemetry.workspace = true
 axum-prometheus.workspace = true
+moka = "0.9"
```
ozwaldorf, line 33: small nit could be workspace dependency

kckeiks:
Numbers improved as expected. On average before adding the cache, the gateway was somewhere around `~3400 req/s` and after adding the cache it was `~3900 req/s`. The indexer was running locally. In production, there could potentially be more latency between indexer and gateway so the difference seen here could possibly be higher in prod.

b0xtch (approved):

PR #390, opened 2023-02-13
feat: run the consensus in main

qti3e:
## Why

Because we need it.

## What

- Loads the Narwhal configuration.
- Runs the Narwhal consensus from ursa binary.

## Checklist


b0xtch (approved):

b0xtch (dismissed):

b0xtch (approved):

crates/ursa-consensus/src/config.rs
```diff
@@ -1,73 +1,155 @@
...
-    #[serde(skip)]
-    keypair: OnceCell<Arc<AuthorityKeyPair>>,
+#[derive(Debug, Clone, Deserialize, Serialize)]
+// When deserializing the config file, use the default from the Default instance
+// to fill any missing field.
+#[serde(default)]
+pub struct ConsensusConfig {
+    /// The address in which the primary will listen for incoming requests on. This MUST
+    /// be a UDP address.
+    pub address: Multiaddr,
```
ozwaldorf, line 22: Should we provide individual field defaults for future upgradability? Mirroring what the other crates provide, for example:
https://github.com/fleek-network/ursa/blob/main/crates/ursa-rpc-service/src/config.rs#L4-L17

ozwaldorf: This can be done separately too, otherwise LGTM

ozwaldorf (approved):

crates/ursa-consensus/src/config.rs
qti3e: we do have defaults, i'm just using the  `#[serde(default)]` over the entire struct which fallbacks to `Default` impl on the type, i think this is cleaner.

PR #392, opened 2023-02-14
chore: change default gateway indexer url

b0xtch:
## Why

- right now we use the dev cid contact so we should use the dev one
- 
## Checklist


ozwaldorf (approved):
mint

PR #393, opened 2023-02-14
chore: ursa main and shutdown controller

qti3e:
## Why

This PR tries to clean up main.rs a little bit and also introduces the shutdown controller, this is the first step in having the `utils` crate, and more code will be moved to the new utils crate eventually.

Also on a logical side, previously services could fail to execute without shutting down the process, which after talking with @b0xtch, we confirmed is not the intended behavior, and the new shutdown controller makes it possible to handle these scenarios.

Fixes # (issue)

## What

- create the new `ursa-utils` crate!
- refactor main.rs
- write a shutdown controller logic
- fix inconsistent sub-service failures not leading to process shutdown.

## Checklist

### DIFF
--- a/crates/ursa-utils/Cargo.toml
+++ b/crates/ursa-utils/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "ursa-utils"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tokio.workspace = true
--- a/crates/ursa-utils/Cargo.toml
+++ b/crates/ursa-utils/Cargo.toml
@@ -1,7 +1,11 @@
 [package]
 name = "ursa-utils"
-version = "0.1.0"
-edition = "2021"
+authors = ["b0xtch <mahmoud@fleek.co>", "Parsa Ghadimi <parsa@fleek.co>"]
+description = "Ursa's utility function"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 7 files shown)


crates/ursa-utils/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "ursa-utils"
+version = "0.1.0"
+edition = "2021"
```
b0xtch, line 4: maintain consistency with other crates

crates/ursa/src/main.rs
```diff
@@ -28,155 +29,181 @@ mod ursa;
...
+        std::process::exit(1);
+    }
+
+    Ok(())
+}
+
+// TODO(qti3e) The current function currently works and it tries to start every service, since
+// each service is from a different crate, the code ownership of this function is unclear, in
+// future each crate should only expose a function that performs everything from their side and
+// main.rs should handle the least amount of logic to actually perform anything.
```
b0xtch, line 55: HUGE, lets extract this into a task

b0xtch (dismissed):

crates/ursa-utils/Cargo.toml
qti3e: nice catch

b0xtch (dismissed):

crates/ursa-utils/src/shutdown.rs
```diff
@@ -0,0 +1,45 @@
+//! The shutdown module makes it easy for us to have one source for sending shutdowns
```
kckeiks, line 1: It might be useful to also add a handler for SIGTERM. What do you think?

kckeiks:
This is a great idea! Looks good. Just one small comment/feature request :)

crates/ursa-utils/src/shutdown.rs
qti3e: Yes! that's a great suggestion. I added SIGTERM and SIGQUIT handlers now.

b0xtch (approved):

ozwaldorf (approved):

PR #400, opened 2023-02-17
feat(node): send cache requests to close peers

### LINKED ISSUE #357: chore: large number of CacheRequests being sent

kckeiks:
## Why

Fixes #357

In testing.

On PUT cache requests tells peers that they should cache the data as well.
I think the reason for doing this is that we want the data to be replicated to other nodes 
so that users closer to those peers can retrieve that same data.

My assumption is that this replication behavior should have a bound because 
it wouldn't be too useful for nodes to tell peers that are far away (node is in San Francisco and peers are in Chicago? There seem to be connections this far apart at the moment) to cache some content. Can we assume that users in far way regions would want that same content most of the time?

Another idea I was toying with is using latency instead of location but latency is a bit volatile. Open to suggestions of course.

## What

- We still have a hash set of peers but in addition, we have a smaller set of peers that are closer to the node.
- Graphsync is now using this subset. Everything else should be intact.
- This functionality is not enabled by default so it doesnt break tests or break local testing.

## Checklist


kckeiks:
@qti3e I know you said you were interested in regionizing the network. Please let me know your thoughts and if there are any ideas here that we could use. 🙏

kckeiks:
Replaced by #427.

PR #401, opened 2023-02-17
chore: tokio-multi-thread-flavor

b0xtch:
## Why

- use the multi-thread flavor of tokio 

## What

- Change most bins to use the multi thread flavor
- 
## Checklist


ozwaldorf:
doesnt the `tokio::main` macro use multi thread by default?

ozwaldorf:
The docs say to just use `tokio::main` for multi thread runtime: https://docs.rs/tokio/latest/tokio/attr.main.html#using-the-multi-thread-
runtime

PR #403, opened 2023-02-19
refactor(config): use serde default on each struct directly

ozwaldorf:
## Why

simplifies default value logic for configs. Functionality should be the exact same

## Checklist

### DIFF
--- a/crates/ursa-application/src/config.rs
+++ b/crates/ursa-application/src/config.rs
@@ -1,22 +1,16 @@
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(default)]
 pub struct ApplicationConfig {
     ///The address the application is listening on. defaults to "0.0.0.0:8003"
-    #[serde(default = "ApplicationConfig::default_domain")]
     pub domain: String,
-}
-
-impl ApplicationConfig {
-    fn default_domain() -> String {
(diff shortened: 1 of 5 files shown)


b0xtch:
I thought we already agreed on which pattern to use?

ozwaldorf:
@b0xtch Didn't realize before that we can just put the serde default attribute directly on the entire struct, previously I thought that it was only available for individual fields. @qti3e brought up that we can use this pattern instead when I asked about it for the ursa-consensus crate's default pattern

b0xtch:
> @b0xtch Didn't realize before that we can just put the serde default attribute directly on the entire struct, previously I thought that it was only available for individual fields. @qti3e brought up that we can use this pattern instead when I asked about it for the ursa-consensus crate's default pattern.
> 
> I also triple checked the functionality is the exact same

I brought it up in this PR https://github.com/fleek-network/ursa/pull/242

ozwaldorf:
Yeah at the time wasn't aware the attribute works on structs directly, the docs only gave examples/details of it used for fields only

b0xtch:
> Yeah at the time wasn't aware the attribute works on structs directly, the docs only gave examples/details of it used for fields only



> Yeah at the time wasn't aware the attribute works on structs directly, the docs only gave examples/details of it used for fields only

But in that PR I explain that, but I remember the decision was to use the function defaults instead as the new pattern.

PR #404, opened 2023-02-20
fix: invalid addresses are being published

kckeiks:
## Why

Invalid addresses are being published because we're mutating the address in the provider engine instead of working with a fresh copy every time we handle it.

```
INFO ursa_index_provider::provider: Announcing the advertisement with the message Message { Cid: Cid(bafy2bzaceaxrvnmnbeelm2vmjlbs7rbin4p4ovfekgezr63kqerixll7jgney), Addrs: ["/ip4/64.225.0.251/tcp/4069/http/p2p/12D3KooWQUYSiMjHVVd6BC2U2dsiaM8YhnfszAF9ruwyp99kym6a/http/p2p/12D3KooWQUYSiMjHVVd6BC2U2dsiaM8YhnfszAF9ruwyp99kym6a"], ExtraData: [] }
at crates/ursa-index-provider/src/provider.rs:185
```

## What

- Clone and pass address.

## Checklist

### DIFF
--- a/crates/ursa-index-provider/src/engine.rs
+++ b/crates/ursa-index-provider/src/engine.rs
@@ -163,7 +163,7 @@ where
                         } else {
                             match self
                                 .provider
-                                .create_announce_message(peer_id, &mut self.addresses)
+                                .create_announce_message(peer_id, self.addresses.clone())
                             {
                                 Ok(announce_message) => {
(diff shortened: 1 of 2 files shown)


b0xtch (approved):
Nice catch

PR #410, opened 2023-02-21
fix the make install command

qti3e:
## Why

supposedly, `cargo install` does not support workspaces properly, so at this point using it leads to different builds from `cargo build` and also different dependency trees, which means our current make file is broken.

Fixes # (issue)

## What

- Remove the use of `cargo install`.
- Replace it with `mkdir -p` and `cp`.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,8 @@ version:
 run: version
 	cargo run --bin cli
 
-install: version
-	cargo install --locked --path crates/ursa --force
+install: version build
+	cp ./target/release/ursa ~/.cargo/bin/
 
 build: version
 	cargo build --release --bin ursa
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ run: version
 	cargo run --bin cli
 
 install: version build
+	mkdir -p ~/.cargo/bin/
 	cp ./target/release/ursa ~/.cargo/bin/
 
 build: version


Makefile
```diff
@@ -10,8 +10,9 @@ version:
 run: version
 	cargo run --bin cli
 
-install: version
-	cargo install --locked --path crates/ursa --force
+install: version build
```
ozwaldorf, line 13: Since we're just building without --locked and manually copying, couldn't we just use install without --locked instead, since that works fine?

qti3e: no `cargo install` just doesn't work, with or without `--locked` 😭

ozwaldorf: oh, i think you're right :cry:

ozwaldorf (approved):

b0xtch:
Just getting to this now; we should prob revert until we fix the documentation and scripts. This is not 1-1 with old `cargo install` as it assumes you have built the project. Open to any plan

ozwaldorf:
> Just getting to this now; we should prob revert until we fix the documentation and scripts. This is not 1-1 with old `cargo install` as it assumes you have built the project. Open to any plan 

just devils advocate since it always runs `version build` before copying it would always succeed

qti3e:
the build does happen, look at line 13.

b0xtch:
That makes sense. It still holds that `cargo install` would fail, hence my comment right? 

I am fine with telling people to use our make file script to install. But if they use `cargo install` it would still fail.

qti3e:
that's right. but the title of the PR is `fix the make install command` not `fix cargo install command`, we still have to figure out how to use the `workspace-hack` properly or what sort of modifications we might need to make the `cargo install` work, but it was not trivial for the moment. i can spend some time again on this task a bit later.

and for the documentation, we're using `make install` as far as i was told, so the motive of the PR was to fix `make install` that is used in documents and Helder's script.

qti3e:
one idea that just crossed my mind right now, is to add the package that is causing problems to our [empty-cargo-workspace](https://github.com/fleek-network/empty-workspace-hack)'s Cargo.toml, i will give it a try later in the day. but i don't see any reason to revert this change just yet. once `cargo install` is fixed. we can revert this commit.

b0xtch:
That shouldn't be the case; tooling should be changed only if we assume everyone will use the Make. For example, the Readme of cli still says to use `cargo install.` So change was unnecessary, and we could've just changed Helders script to do this.

b0xtch:
Given that there are plans to revert to this, also our CI doesn't use any make commands

b0xtch:
We don't have to revert this now, just for future reference

PR #413, opened 2023-02-22
feat: ABCI engine

daltoncoder:
## Why

This PR integrates the ABCI interface as the engine that drives the application layer. It also introduces a rpc shim to forwards txns to consensus, or querys directly to the application layer. After a txn goes to consensus the certificate is forwarded to the application layer and executed. All txns going through N/B before being executed will maintain equal state replication amongst our nodes. This shim is temporary and the next steps would be to integrate this functionality into the ursa_rpc_service. There is a genesis file that loads staking, token, and nodeRegistry contract into the application layer at block 0.

Some other next steps would be to start utilizing the nodeRegistry contract and pulling our committee data from there. 

Fixes # (issue)

## What
- Added ABCI engine
- All narwhal batches are forwarded to this engine and sent to be executed on application layer

## Notes

- included a bin folder with a separate bin that can be used to test the contracts and state replication. 

## Checklist


crates/ursa-application/README.md
```diff
@@ -0,0 +1,3 @@
+# Ursa application layer
+
+> Ursa's evm based application layer. Bullshark certificates contain txns are forwarded to this layer via an ABCI engine and executed in the vm.
```
b0xtch, line 3:
```suggestion
> Ursa's EVM-based application layer. Bullshark certificates containing transactions are forwarded to this layer via an ABCI engine and executed in the VM. 
```

crates/ursa-consensus/src/server.rs
```diff
@@ -0,0 +1,84 @@
+use crate::{AbciQueryQuery, BroadcastTxQuery};
+use multiaddr::{Multiaddr, Protocol};
+use narwhal_types::{TransactionProto, TransactionsClient};
+use tendermint_proto::abci::ResponseQuery;
+use tokio::sync::mpsc::Sender;
+use tokio::sync::oneshot::{channel as oneshot_channel, Sender as OneShotSender};
+use tracing::{debug, error};
+use warp::{Filter, Rejection, Reply};
```
b0xtch, line 8: future: when this moves to the rpc service, we use axum

crates/ursa/src/main.rs
```diff
@@ -175,12 +179,45 @@ async fn run() -> Result<()> {
         }
     });
 
+    // Start the application server
```
b0xtch, line 182: floating comment

crates/ursa-consensus/src/abci_engine.rs
```diff
@@ -0,0 +1,189 @@
...
+
+pub struct Engine {
+    /// The address of the ABCI app
+    pub app_address: SocketAddr,
+    /// Messages received from the ABCI Server to be forwarded to the engine.
+    pub rx_abci_queries: Receiver<(OneShotSender<ResponseQuery>, AbciQueryQuery)>,
+    /// The last block height, initialized to the application's latest block by default
+    pub last_block_height: i64,
+    pub client: AbciClient,
+    pub req_client: AbciClient,
```
b0xtch, line 26: Whats the difference between these two?

crates/ursa-consensus/src/abci_engine.rs
```diff
@@ -0,0 +1,189 @@
...
+
+// Tendermint Lifecycle Helpers
+impl Engine {
+    /// Calls the `InitChain` hook on the app, ignores "already initialized" errors.
+    pub fn init_chain(&mut self) -> Result<()> {
+        let mut client = ClientBuilder::default().connect(self.app_address)?;
+        match client.init_chain(RequestInitChain::default()) {
+            Ok(_) => {}
+            Err(err) => {
+                tracing::error!("{:?}", err);
```
b0xtch, line 139:
```suggestion
                error!("{:?}", err);
```

crates/ursa-consensus/src/abci_engine.rs
```diff
@@ -0,0 +1,189 @@
...
+                // ignore errors about the chain being uninitialized
+                if err.to_string().contains("already initialized") {
+                    warn!("{}", err);
+                    return Ok(());
+                }
+                bail!(err)
+            }
+        };
```
b0xtch, line 147:
```suggestion
        if let Err(err) = client.init_chain(RequestInitChain::default()) {
            error!("{:?}", err);
            // ignore errors about the chain being uninitialized
            if err.to_string().contains("already initialized") {
                warn!("{}", err);
                return Ok(());
            }
            bail!(err)
        };
```

crates/ursa-consensus/src/server.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// Simple HTTP API server which listens to messages on:
+/// * `broadcast_tx`: forwards them to Narwhal's mempool/worker socket, which will proceed to put
+/// it in the consensus process and eventually forward it to the application.
+/// * `abci_query`: forwards them over a channel to a handler (typically the application).
+pub struct AbciApi<T> {
+    mempool_address: String,
```
b0xtch, line 15:
```suggestion
    mempool_address: Multiaddr,
```

crates/ursa-consensus/src/server.rs
```diff
@@ -0,0 +1,84 @@
...
+    ) -> Self {
+        let mempool_port = mempool_address
+            .iter()
+            .find_map(|proto| match proto {
+                Protocol::Tcp(port) => Some(port),
+                _ => None,
+            })
+            .expect("Expected tcp url for worker mempool");
+
+        let mempool_address = format!("http://0.0.0.0:{}", mempool_port);
```
b0xtch, line 32: Is this a re-instantiation of the variable `mempool_address`? If so, could we rename it for readability? its based into the new function as type  `Multiaddr` and stored as type `String`

crates/ursa-consensus/src/abci_engine.rs
daltoncoder: Queries will go through the req_client because they dont change state. Txns will go through client. The reason for having separate clients for this is because the client will maintain a lock while executing or committing transactions on the evm db. So having separate channels will not hold up queries when this is happening, and queries will return the last committed state, and not state in an inbetween state if its half way through executing.

b0xtch: Having separate clients isn't necessary to separate reads from writes. Look at this client by aleo https://github.com/AleoHQ/aleo_lambda_blockchain/blob/main/src/client/tendermint.rs

crates/ursa-consensus/src/server.rs
daltoncoder: Narwhals TransactionClient doesnt support Multiaddr so if we change it on the struct we would have to do the conversation every time we receive a request instead of once in the new() function

crates/ursa-consensus/src/server.rs
```diff
@@ -0,0 +1,84 @@
...
+/// Simple HTTP API server which listens to messages on:
+/// * `broadcast_tx`: forwards them to Narwhal's mempool/worker socket, which will proceed to put
+/// it in the consensus process and eventually forward it to the application.
+/// * `abci_query`: forwards them over a channel to a handler (typically the application).
+pub struct AbciApi<T> {
+    mempool_address: String,
+    tx: Sender<(OneShotSender<T>, AbciQueryQuery)>,
+}
+
+impl<T: Send + Sync + std::fmt::Debug> AbciApi<T> {
```
b0xtch, line 19:
```suggestion
impl<T: Send + Sync + Debug> AbciApi<T> {
```

b0xtch (dismissed):

b0xtch (approved):

PR #418, opened 2023-03-01
fix(full-node): add a 72hr max retention time for tsdb (prom)

ozwaldorf:
## Why

Prometheus is taking up a large amount of space due to uncapped retention time. Sets a generous limit at 72h

## Notes

We could do less, say 24hrs instead

## Checklist

- [x] ~~I have run the app using my feature and ensured that no functionality is broken~~


matthias-wright (approved):
lgtm

PR #427, opened 2023-03-02
feat(node): send cache requests to peers in replication set

kckeiks:
## Why

Fixes https://github.com/fleek-network/ursa/issues/357

Too many cache requests are being sent.

## What

- Add a manager to handle access to list of connected peers and a separate subset of that list (replication set).
- When we receive rtt from a PING, we pass the event to manager to update the replication set.
- Our assumption of rtt is that if it's high (higher than some threshold), then the peer must be far away.  High rtt could also indicate network congestion. Thus, peers will not be added to the replication set if rtt is high because we don't want to use this set to replicate content to peers who are far away and/or make network congestion worse. There are other factors that affect rtt but for the problem at hand #357, these assumptions seem reasonable. Open to suggestions of course.

## Checklist

### DIFF
--- a/crates/ursa-network/src/behaviour.rs
+++ b/crates/ursa-network/src/behaviour.rs
@@ -43,15 +43,16 @@ use libp2p::{
 };
 use libp2p_bitswap::{Bitswap, BitswapConfig};
 use std::borrow::Cow;
+use std::iter;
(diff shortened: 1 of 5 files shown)


b0xtch (approved):
Mint

PR #430, opened 2023-03-08
chore: 🤖 curl is required by health check

heldrida:
## Why

Curl is required by health check

Fixes # (issue)

## What

- Add dependency in apt-get command

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ FROM debian:bullseye-slim
 
 RUN apt-get update && apt-get install -y \
     libcurl4-openssl-dev \
+    curl \
     && apt-get clean \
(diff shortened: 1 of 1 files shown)


matthias-wright (approved):

Dockerfile
```diff
@@ -24,6 +24,7 @@ FROM debian:bullseye-slim
 
 RUN apt-get update && apt-get install -y \
     libcurl4-openssl-dev \
+    curl \
```
ozwaldorf, line 27: nitpick
```suggestion
    libcurl4-openssl-dev curl \
```

ozwaldorf (approved):

PR #441, opened 2023-03-14
Fix/remove coverage ci

daltoncoder:
## Why

We are failing on github CI. This is may be a better solution then #440 since limiting job makes the task take 68 minutes

## Checklist


b0xtch (approved):
lets link issue

PR #447, opened 2023-03-22
chore(infra/tf): use images instead of building + maintenance

ozwaldorf:
## Why

For `infra/tf`, we can use our released docker images instead of building fresh every time. There are also some other small issues needing cleanup there.

## What

- updated bootstrap node
  - deploy time reduced from ~30mins -> less than 5mins
  - remove docker install stuff since we deploy a base image with it preinstalled
  - sets up `~/.ursa/config.toml`
  - deploys docker/full-node, minus nginx

- updated provider node
  - deploy time reduced from ~30mins -> less than 5mins
  - provider has it's own cloud conf, stripped of the docker install stuff
  - sets up `~/.ursa/config.toml` with the node ip for proper advertisements
  - sets up ssl with the domain name
 
- updated metrics
  - maxminddb is automated

- updated firewall
  - opened port 3000 for access to grafana
  - opened p2p port 4890 to allow quic connections
 
- added Makefile scripts for infra/tf
  - init-tf, plan-tf, apply-tf, destroy-tf
  - nodes-tf: list node ips
  - ping-tf: pings nodes to see if ursa is up
  - exec-tf: executes a command on each provider node

- updated readme

## Notes

We also migrated to a new digitalocean account which doesn't have the same domain setup, so I swtiched the default to `ursa.zone`

## Checklist


b0xtch:
What you mean by “didn't work for deploying testnet nodes quickly”.

b0xtch:
> > What you mean by “didn't work for deploying testnet nodes quickly”.
> 
> 
> 
> Since we're switching to the new DO, minor modifications were required to deploy. Also the cloud conf for nodes was set to `bootstrap_node.yml`, and was a mess of commented out code which didn't deploy the node. Now it properly deploys bootstraps as bootstraps and nodes as nodes
> 
> 
> 
> > For the firewall, only 6009 and 80/443 should be exposed
> 
> 
> 
> We have a p2p port 4890 for quic that should be public, and It'd be nice to be able to see the node metrics via grafana on 3000.

But you just commented the mess, how is it faster?

b0xtch:
> > But you just in in uncommented the mess, how is it faster?
> 
> 
> 
> It's completely different now. All the extra docker install steps are removed since we deploy a base image with docker already installed. Then it modifies the compose to use the latest docker release of ursa, so increasing speed by removing build time. For bootstrap nodes, there's also an extra step to disable nginx since they don't use it. Nodes are up and responding to /ping in less than 5 minutes now vs 30+ mins

I would modify the description of the PR as it's not accurate

infra/tf/main.tf
```diff
@@ -96,13 +112,23 @@ resource "digitalocean_firewall" "ursa-network" {
...
+  inbound_rule {
+    port_range       = "6009"
+    protocol         = "tcp"
+    source_addresses = ["0.0.0.0/0", "::/0"]
+  }
+  inbound_rule {
+    port_range       = "4890"
+    protocol         = "udp"
     source_addresses = ["0.0.0.0/0", "::/0"]
   }
```
b0xtch, line 133: 6009 and 80/443 should only be exposed. In this case only 6009 for ursa network

infra/tf/provider.tf
```diff
@@ -12,5 +12,5 @@ provider "digitalocean" {
 }
 
 data "digitalocean_ssh_key" "ursa-dev" {
-  name = "ursa-dev"
+  name = "ossian"
```
b0xtch, line 15:
```suggestion
  name = "ursa-dev"
```

infra/tf/configs/provider_node.yml
```diff
@@ -0,0 +1,31 @@
+#cloud-config
+groups:
+  - ubuntu: [root,sys]
+
+runcmd:
```
b0xtch, line 5: could we use `get.fleek.network` script instead

b0xtch:
> > I would modify the description of the PR as it's not accurate
> 
> Agreed it could've been better, amended

I believe the description is still inaccurate, `use images instead of building`

infra/tf/configs/provider_node.yml
ozwaldorf: It's inconsistent for automation purposes in it's current state, I wanted to use it but ran into some issues and at least atm, it would require updating inputs the cloud-init gives if it changes at all. 

If we add like a `--noninteractive` mode and a way to pass in the domain and stuff with no user input, it could be used.

For bootstrap nodes though, we'd need a way to skip nginx steps in the assisted installer

infra/tf/main.tf
ozwaldorf: What about our public quic port 4890 for libp2p? Enabled by default on all ursa nodes

And 3000 would be nice to be able to get to the grafana dashboard to watch the node health

Will remove 4069 though had it on for testing

infra/tf/configs/provider_node.yml
ozwaldorf: cc @heldrida

infra/tf/main.tf
b0xtch: yeah, 4069 shouldn't be exposed,  4890 yeah that could stay, everything else is a balance. I suppose a metrics port is fine

b0xtch (dismissed):

infra/tf/main.tf
```diff
@@ -135,20 +161,24 @@ resource "digitalocean_record" "ursa-dashboard-domain" {
 
 resource "digitalocean_droplet" "ursa-dashboard" {
   image      = var.droplet_image
-  name       = "dashboard"
+  name       = "ursa-dashboard"
   region     = var.droplet_region
-  size       = var.droplet_size
+  size       = "s-8vcpu-16gb-intel"
```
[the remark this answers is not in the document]
ozwaldorf: Since there's no response, I'll regard that my proposed solution is acceptable
https://github.com/fleek-network/ursa/pull/447/commits/4377918ebed836e529d1d27ba53c9adcfb975481

infra/tf/configs/provider_node.yml
ozwaldorf: Is this a blocker for the pr?

b0xtch:
What’s the status of this PR?

PR #454, opened 2023-03-28
feat: dynamic epoch+committee

daltoncoder:
## Why

Our Consensus is designed to work within epochs. After a set amount of time an epoch should end, a new committee should be decided on randomly from all the nodes in the network and restart itself with the new epoch parameters and committee members. Before we were just running a fixed version of narwhal from the config.

This PR adds a wrapper around our Narwhal(consensus.rs). This wrapper will pull epoch/committee information from application layer before starting narwhal and start narwhal with the right committee and parameters. This wrapper will also send a request when the epoch end time has passed to signal the node is ready to move epochs.

The epoch change works like this: Every node in the committee keeps a timer to epoch end, when that timer passes it sends a certificate to be ordered that says it is ready to change but stays in the current epoch. When 2/3rds of the nodes in the committee have sent a certificate saying they are ready to move epochs, that last certificate will trigger the epoch change for all nodes. This way we ensure all nodes move to the next epoch at the same exact point.

Fixes # (issue) - No issue was created for this but I will be better with this going forward

## What

- Renamed service.rs to narwhal.rs. Added a wrapper around it in consensus.rs that is in charge of starting the entire consensus service.
- If a node is not on the committee it will no longer run narwhal at all. 
- We no longer need a committee.json for narwhal. Everything should be in the genesis.toml in ursa-application.
- Some utils like like load_json where moved to ursa_utils crate from consensus
- Introduced a forge solidity repo into the mono repo in the contracts folder. Has the contracts we are loading in the genesis.toml.
- Since epoch starts based on the params of the genesis file i made a simple genesis script that will update epoch 0 to start at the time the script was ran, the script also takes an optional parameter of epoch length in milliseconds(defaults to 5 minutes) `cargo run --bin genesis 600000` would update the genesis to start now with an epoch time of 10 min. This script also sets genesis committee to the nodes from https://github.com/qti3e/ursa-nodes for easy local testing and running of the committee.

## Checklist


contracts/src/Epoch.sol
```diff
@@ -0,0 +1,140 @@
...
+
+    NodeRegistry public nodeRegistry;
+    bool private initialized;
+    uint256 private readyToChange;
+
+    struct CommitteeMember {
+        string publicKey;
+        string primaryAddress;
+        string workerAddress;
+        string workerMempool;
```
b0xtch, line 29: isn't this a list of worker addresses?

contracts/src/Epoch.sol
```diff
@@ -0,0 +1,140 @@
...
+
+    function _changeEpoch() private {
+        epoch++;
+
+        committee[epoch] = _chooseNewCommittee();
+        currentEpochEndStampMs = currentEpochEndStampMs + epochDurationMs;
+        currentCommitteeSize = committee[epoch].length;
+    }
+
+    function _chooseNewCommittee() private view returns (string[] memory _committee) {
```
b0xtch, line 115: is the new committee drive by the application layer?

contracts/src/FLK.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: UNLICENSED
```
b0xtch, line 1: same here

contracts/src/FLK.sol
```diff
@@ -0,0 +1,82 @@
...
+pragma solidity ^0.8.15;
+
+import "solmate/tokens/ERC20.sol";
+
+contract FleekToken is ERC20 {
```
b0xtch, line 6: Should we run this under a proxy contract?

crates/ursa-consensus/src/consensus.rs
```diff
@@ -0,0 +1,272 @@
...
+            }
+        }
+        self.start_current_epoch().await
+    }
+
+    async fn wait_to_signal_epoch_change(&self, time_until_change: Duration) {
+        let primary_public_key = self.narwhal_args.primary_keypair.public().clone();
+        let mempool_address = self.mempool_address.clone();
+        tokio::task::spawn(async move {
+            tokio::time::sleep(time_until_change).await;
```
b0xtch, line 170:
```suggestion
            time::sleep(time_until_change).await;
```

crates/ursa-consensus/src/consensus.rs
```diff
@@ -0,0 +1,272 @@
...
+            // We shouldnt panic here lets repeatedly try
+            loop {
+                tokio::time::sleep(Duration::from_secs(1)).await;
+                let txn = match build_transaction(
+                    EPOCH_ADDRESS,
+                    "signalEpochChange(string memory committeeMember):(bool)",
```
b0xtch, line 176: This ABI could be structs or types or consts

crates/ursa-consensus/src/narwhal.rs
```diff
@@ -54,33 +47,32 @@ enum Status {
...
-            true,
-            arguments.registry_service.clone(),
-        );
+impl NarwhalService {
+    /// Create a new narwhal service using the provided arguments.
+    pub fn new(
+        arguments: NarwhalArgs,
+        store: NodeStorage,
+        committee: Arc<ArcSwap<Committee>>,
+        worker_cache: Arc<ArcSwap<WorkerCache>>,
```
b0xtch, line 56: Pull main, we no longer use ArcSwap

crates/ursa-utils/src/contract_bindings/epoch.rs
```diff
@@ -0,0 +1,652 @@
+pub use epoch::*;
```
b0xtch, line 1: Whats this file

crates/ursa-utils/src/contract_bindings/node_registry.rs
```diff
@@ -0,0 +1,524 @@
+pub use node_registry::*;
```
b0xtch, line 1: Whats this file?

crates/ursa-utils/src/transactions.rs
```diff
@@ -75,6 +81,103 @@ pub fn encode_params(func: &Function, args: &[impl AsRef<str>]) -> Result<Vec<u8
     }
 }
 
+pub fn get_epoch_info_params() -> TransactionRequest {
+    // safe unwrap
+    let to = "0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAC"
```
b0xtch, line 86: make const

crates/ursa-consensus/src/abci_engine.rs
```diff
@@ -1,62 +1,80 @@
...
         }
     }
 
-    /// Receives an ordered list of certificates and apply any application-specific logic.
-    pub async fn run(&mut self, mut rx_output: Receiver<Vec<Batch>>) -> Result<()> {
+    pub async fn start(&mut self) -> Result<()> {
         self.init_chain()?;
 
         loop {
             tokio::select! {
```
b0xtch, line 76:
```suggestion
            select! {
```

crates/ursa-consensus/src/consensus.rs
```diff
@@ -0,0 +1,272 @@
...
+                state.narwhal.shutdown().await
+            }
+        }
+        self.start_current_epoch().await
+    }
+
+    async fn wait_to_signal_epoch_change(&self, time_until_change: Duration) {
+        let primary_public_key = self.narwhal_args.primary_keypair.public().clone();
+        let mempool_address = self.mempool_address.clone();
+        tokio::task::spawn(async move {
```
b0xtch, line 169:
```suggestion
        task::spawn(async move {
```

crates/ursa-consensus/src/consensus.rs
```diff
@@ -0,0 +1,272 @@
...
+            tokio::time::sleep(time_until_change).await;
+            // We shouldnt panic here lets repeatedly try
+            loop {
+                tokio::time::sleep(Duration::from_secs(1)).await;
```
b0xtch, line 173:
```suggestion
                time::sleep(Duration::from_secs(1)).await;
```

crates/ursa-consensus/src/consensus.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    pub async fn start(&mut self) {
+        self.start_current_epoch().await;
+        loop {
+            let reconfigure_future = self.reconfigure_notify.notified();
+            let shutdown_future = self.shutdown_notify.notified();
+            tokio::pin!(shutdown_future);
+            tokio::pin!(reconfigure_future);
+            tokio::select! {
```
b0xtch, line 216:
```suggestion
            pin!(shutdown_future);
            pin!(reconfigure_future);
            select! {
```

b0xtch (commented):
Big Nit: 

```rust
///Comment
vs

/// Comment
```

did a quick run through as its in draft

contracts/src/Epoch.sol
```diff
@@ -0,0 +1,140 @@
+// SPDX-License-Identifier: UNLICENSED
```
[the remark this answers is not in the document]
daltoncoder: Added but i cant put multiple like we have  in other files `// SPDX-License-Identifier: Apache-2.0, MIT`
Solidity compiler actually throws an error. Put `// SPDX-License-Identifier: MIT` for now.

There have been discussions to remove this error from compiler but they decided to keep it for reasons idk off the top of my head

contracts/src/Epoch.sol
daltoncoder: It can be but we only have 1 worker per node for now so I stored it like this. Ill refactor to future proof it

contracts/src/FLK.sol
daltoncoder: We should, but implementation contract shouldnt really be aware its under a proxy in its code. Ive been thinking Proxy/umbrella/governance contract that can manage these as future work

crates/ursa-utils/src/contract_bindings/epoch.rs
daltoncoder: This is autogenerated file from ethers that handles contract bindings. Very useful for taking bytes returned from evm and turning into rust structs/ vice versa. I can extract just the stuff we are using currently to clean it up a bit if you want

crates/ursa-utils/src/contract_bindings/node_registry.rs
daltoncoder: same as above

contracts/src/Epoch.sol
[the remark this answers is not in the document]
b0xtch: Yeah one or the other fine

contracts/src/FLK.sol
b0xtch: Sounds good, as long as we are thinking about upgradability

contracts/src/Epoch.sol
daltoncoder: refactored

crates/ursa-application/src/types.rs
```diff
@@ -119,15 +119,18 @@ impl<Db: AbciDb> ConsensusTrait for Consensus<Db> {
...
         let registry_bytes = hex::decode(genesis.registry.bytecode).unwrap();
+        let epoch_bytes = hex::decode(genesis.epoch.bytecode).unwrap();
         let hello_bytes = hex::decode(genesis.hello.bytecode).unwrap();
 
         //Parse addresses for contracts
         let token_address: Address = genesis.token.address.parse().unwrap();
         let staking_address: Address = genesis.staking.address.parse().unwrap();
         let registry_address: Address = genesis.registry.address.parse().unwrap();
+        let epoch_address: Address = genesis.epoch.address.parse().unwrap();
         let owner_address: Address = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
```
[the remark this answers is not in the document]
daltoncoder: removed, left over from some manual testing

contracts/src/Epoch.sol
daltoncoder: Yes the new committee is decided by the application layer. But the process is only triggered by Narwhal certificates. The random function that selects the new committee is a placeholder for now. A truly random threshold signature that all nodes have a chance to contribute too is the end goal here.

crates/ursa-application/src/bin/genesis_committee.toml
```diff
@@ -0,0 +1,32 @@
+[[committee]]
+    owner = "0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC"
```
kckeiks, line 2: Are these test committees?

crates/ursa-application/src/types.rs
```diff
@@ -245,18 +216,42 @@ impl<Db: AbciDb> ConsensusTrait for Consensus<Db> {
...
         let result = state.execute(tx, false).await.unwrap();
         tracing::trace!("executed tx");
 
+        if to_epoch_contract {
+            if let ExecutionResult::Success {
+                output: Output::Call(bytes),
+                ..
+            } = &result
+            {
+                //  GetCurrentEpochInfoReturn::decode(Bytes::from(output)).unwrap()
```
kckeiks, line 241:
```suggestion
                
```

crates/ursa-consensus/src/abci_engine.rs
```diff
@@ -1,76 +1,111 @@
...
+use std::sync::Arc;
+use tokio::sync::{mpsc, oneshot, Notify};
+use tokio::{pin, select, time};
+use tracing::warn;
+use ursa_utils::shutdown::ShutdownController;
 
 use narwhal_types::{Batch, Transaction};
 
-// Tendermint Types
+//Tendermint Types
```
b0xtch, line 11:
```suggestion
// Tendermint Types
```

crates/ursa-consensus/src/consensus.rs
```diff
@@ -0,0 +1,268 @@
...
+                }
+            }
+        }
+    }
+
+    pub async fn shutdown(&mut self) {
+        self.shutdown_notify.notify_waiters();
+    }
+
+    pub async fn run_edge_node(&self) {
```
b0xtch, line 230: The ursa node would be run from here?

kckeiks (dismissed):
Looks good.

Nit:

`// Doc or comment format consistency, preferably capitalized and end should be marked with a period.`

crates/ursa-consensus/src/consensus.rs
daltoncoder: Maybe its bad naming and its pretty much a placeholder for now. but this is the entry to non validator logic that wouldnt be running narwhal.

contracts/src/FLK.sol
```diff
@@ -0,0 +1,82 @@
...
+        _removeMinter(_account);
+    }
+
+    /**
+     * @dev Mint new tokens.
+     * @param _to Address to send the newly minted tokens
+     * @param _amount Amount of tokens to mint
+     */
+    function mint(address _to, uint256 _amount) external {
+        _mint(_to, _amount);
```
matthias-wright, line 44:
```suggestion
        require(isMinter(_to), "Only minters can mint new tokens");
        _mint(_to, _amount);
```
Should we check if the receiving address is a minter or is that not necessary?

contracts/src/NodeRegistry.sol
```diff
@@ -0,0 +1,106 @@
...
+        linkedListHead = _node.primaryPublicKey;
+    }
+
+    function removeNode(string calldata _nodeAddress) external {
+        _removeNode(_nodeAddress);
+    }
+
+    function _removeNode(string calldata _nodePublicKey) private {
+        whitelist[whitelist[_nodePublicKey].previous] = whitelist[whitelist[_nodePublicKey].next];
+        whitelist[whitelist[_nodePublicKey].next] = whitelist[whitelist[_nodePublicKey].previous];
```
matthias-wright, line 79:
```suggestion
        whitelist[whitelist[_nodePublicKey].previous].next = whitelist[whitelist[_nodePublicKey].next].primaryAddress;
        whitelist[whitelist[_nodePublicKey].next].previous = whitelist[whitelist[_nodePublicKey].previous].primaryAddress;
```
I'm probably missing something obvious here, but why aren't we doing this?
Setting the next pointer of node_to_delete.previous to node_to_delete.next and also setting the previous pointer of node_to_delete.next to node_to_delete.previous.
Also, is it necessary to check for edge cases like the list is empty or only contains one node?

daltoncoder: Good catch this logic was wrong i fixed. Edge cases should be covered since every value is default 0 address in solidity but we need tests on this, which will be in the next PR that gets these contracts production ready.

contracts/src/FLK.sol
daltoncoder: Removed access control from this since this contract isnt integrated into protocol at all yet, and pending transaction verification spec. Was put in for beards tests. Might be worth keeping out of this PR in this state though.

crates/ursa-application/src/types.rs
```diff
@@ -9,9 +9,8 @@ use abci::{
     types::*,
```
kckeiks, line 9: It would be useful to review the unwraps used in this crate, and maybe use Result instead, since one of these methods turned out to be fallible.

b0xtch: anyhow::Result would be a great addition

crates/ursa-application/src/bin/genesis_committee.toml
b0xtch: Could we move `https://github.com/qti3e/ursa-nodes` under the repo

daltoncoder: We could for sure, but its a very manual way of running a whole committee and not very maintainable. I think we ideally need something like this that facebook did in their implementation for running a whole committee https://github.com/facebookresearch/narwhal/tree/main/benchmark
Building something like that is on my roadmap- E2E Benchmarks/simulations. 
@qti3e thoughts on bringing in your repo?

qti3e (changes requested):
The code doesn't look bad on a line-by-line base; most of the lines are doing what is intended to do.

---- Side note:

However, for history-keeping reasons, I also have to say here, probably one last time, that the idea of having EVM as our application layer stinks; we do not run arbitrary ad-hoc logic, which would require redeploying smart contracts (we can always upgrade the node), and even if the choice of a VM would make sense due to deploy on-demand benefit over pure Rust, EVM would still not be the best option to go.

The code already has introduced a ton of inefficiency and reduction in the performance of the reads from the EVM (and its blocking nature), is not fit for a highly available product such as a CDN, and would demand us to have a secondary lock-free in-memory cache of the entire balance map (just an example) in the code base, introducing yet another layer of indirection to the architecture.

The idea that we have to go through a local connection to ourselves to get to the data that is already supposed to be in our memory is just... _for lack of a better word_ bad design.

But apart from the general talk about the entire idea, which I only mentioned, there is a record of me somewhere that I have said it; the implementation of the idea is good, and kudos to @daltoncoder.

P.S The other reason I chose this PR to put up is that this is the first PR trying to merge Solidity source code into the main. However, I do not expect any changes here because, as far as i am concerned, this is not my place to make this decision. And this has been discussed before I was part of the team anyway. I do not expect any changes. I just wanted to have a publicly available note about my opinion about this

crates/ursa-utils/src/contract_bindings/epoch.rs
daltoncoder: Removed and replaced the auto generated files

contracts/src/NodeRegistry.sol
```diff
@@ -0,0 +1,106 @@
...
+        initialized = true;
+    }
+
+    function registerNode(NodeInfo memory _node) external {
+        require(whitelist[_node.primaryPublicKey].owner == address(0), "This node is already on whitelist");
+        _registerNode(_node);
+    }
+
+    function _registerNode(NodeInfo memory _node) private {
+        whitelist[linkedListHead].previous = _node.primaryAddress;
```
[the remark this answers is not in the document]
daltoncoder: Mappings in Solidity are kind of like hash tables, and when it is instantiated all possible keys map to a value whose byte representation is 0 bytes. Same with variables, will start as default value if not assigned when created. Accessing the items before inserting anything is okay as well, you will just be operating on default value. 

So this is actually working as intended, if linked list head is unassigned(0) the new node is top of list and should put its next field as 0

crates/ursa-consensus/src/consensus.rs
```diff
@@ -0,0 +1,257 @@
...
+            .authorities
+            .contains_key(self.narwhal_args.primary_keypair.public())
+        {
+            self.run_edge_node().await;
+            return;
+        }
+
+        // Make or open store specific to current epoch.
+        let mut store_path = self.store_path.clone();
+        store_path.set_file_name(format!("{}-{}", STORE_NAME, epoch));
```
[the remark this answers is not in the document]
daltoncoder: Good idea added the narwhal directory

contracts/src/NodeRegistry.sol
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    function removeNode(string calldata _nodeAddress) external {
+        _removeNode(_nodeAddress);
+    }
+
+    function _removeNode(string calldata _nodePublicKey) private {
+        whitelist[whitelist[_nodePublicKey].previous].next = whitelist[_nodePublicKey].next;
+        whitelist[whitelist[_nodePublicKey].next].previous = whitelist[_nodePublicKey].previous;
+        whitelist[_nodePublicKey].owner = address(0);
```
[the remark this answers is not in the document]
daltoncoder: Added require to check if node is whitelisted that way our whitelist count stays accurate.  My comment above applies to 2 and 3 here

contracts/src/Epoch.sol
```diff
@@ -0,0 +1,136 @@
...
+            if (i == currentCommitteeSize) {
+                revert();
+            }
+        }
+        uint256 roundUp;
+        if (currentCommitteeSize * 2 / 3 > 0) {
+            roundUp = 1;
+        }
+
+        if (readyToChange >= (currentCommitteeSize * 2 / 3) + roundUp) {
```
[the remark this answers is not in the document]
daltoncoder: Good catch

contracts/src/Epoch.sol
```diff
@@ -0,0 +1,136 @@
...
+                readyToChange++;
+                break;
+            }
+            unchecked {
+                i++;
+            }
+            if (i == currentCommitteeSize) {
+                revert();
+            }
+        }
```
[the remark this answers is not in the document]
daltoncoder: We can add a double mapping here and prevent having to iter. I was initially thinking the loop once every 24 hours would be better then the extra storage. Lets optimize this in next PR that brings these contracts production ready. There is a lot to optimize still in them

daltoncoder:
> The code doesn't look bad on a line-by-line base; most of the lines are doing what is intended to do.
> 
> ---- Side note:
> 
> However, for history-keeping reasons, I also have to say here, probably one last time, that the idea of having EVM as our application layer stinks; we do not run arbitrary ad-hoc logic, which would require redeploying smart contracts (we can always upgrade the node), and even if the choice of a VM would make sense due to deploy on-demand benefit over pure Rust, EVM would still not be the best option to go.
> 
> The code already has introduced a ton of inefficiency and reduction in the performance of the reads from the EVM (and its blocking nature), is not fit for a highly available product such as a CDN, and would demand us to have a secondary lock-free in-memory cache of the entire balance map (just an example) in the code base, introducing yet another layer of indirection to the architecture.
> 
> The idea that we have to go through a local connection to ourselves to get to the data that is already supposed to be in our memory is just... _for lack of a better word_ bad design.
> 
> But apart from the general talk about the entire idea, which I only mentioned, there is a record of me somewhere that I have said it; the implementation of the idea is good, and kudos to @daltoncoder.
> 
> P.S The other reason I chose this PR to put up is that this is the first PR trying to merge Solidity source code into the main. However, I do not expect any changes here because, as far as i am concerned, this is not my place to make this decision. And this has been discussed before I was part of the team anyway. I do not expect any changes. I just wanted to have a publicly available note about my opinion about this

Thanks for the review and using my PR to state your sidepoints. I wont get into what i agree or disagree with here, its not really the place. We have multiple calls a week to discuss protocol design, which is how we ended up with these decisions.

crates/ursa-application/src/bin/genesis_committee.toml
b0xtch: I think if it will be used then it should be moved over. If alternatives exist like the ones mentioned, that makes it better too

crates/ursa-application/src/types.rs
daltoncoder: Cleaned up the init_chain function a bit. If we cant load genesis we should panic and loading in static bytes should not ever, but cant return anyhow::Result in this particular function because its part of the abci spec. Genesis refactor on the roadmap the plan is to move loading genesis out of this function and load it directly into parity db on the app creation, then we can switch it up to anyhow.

qti3e (approved):

b0xtch (approved):
Noice, this is moving the needle in the right direction. In terms of follow-ups and what's next for our Actor model, we can talk about it. Parallel work can coexist to prevent any bottlenecks. There could be a case where we play the FVM model Rust and EVM support. 

We have to start somewhere, and this is a good start.

Let's add the feedback from this PR to the roadmap

Issue #452, opened 2023-03-27
feat: metric estimations


matthias-wright:
Nodes should estimate metrics such as bandwidth, latency, and uptime for the nodes they interact with.

PR #456, opened 2023-03-28
feat: bandwidth measurements

matthias-wright:
## Why
We need measurements for the metrics that will be used for the reputation system, see #452. This PR starts with basic bandwidth measurements. 

## What
- Added `MeasurementManager` where we can register requests and responses for each peer. 
- The manager will compute bandwidth estimates based on the size of data that was transferred and the duration (see diagram below).
- For now, bandwidth won't be estimated for upload and download separately. We estimate the average of the upload and download bandwidth.
- This PR also introduces a new dependency: [lru-rs](https://github.com/jeromefroe/lru-rs). I'm aware that we already using `moka` which could be used here as well for the cache, but I ran a quick benchmark and `lru-rs` seems faster for this case (likely because it's not thread-safe, which we don't need here atm).

## Notes
bandwidth_estimate
- For the diagram above, the bandwidth will be estimated as follows:
`bandwidth = (size(payload1) + size(payload2)) / (t4 - t1)`
- We have to make the assumption that `t3 - t2` is relatively small.

## Future Work
- Estimates for latency and uptime.
- Some improvements like discarding measurements where the duration or the data size is too small, because they would likely skew the average.
- Alternatively, weighting the average by the duration or the file size of the estimate (longer duration / larger file size probably implies a more accurate estimate).

## Checklist


matthias-wright:
Thanks for the feedback. `Bandwidth` handles the bandwidth estimates for a specific peer. The current estimate can be obtained using `get_estimate`.
The reason for the two fields `sum` and `count` is that we are calculating the average over a stream of values. Each request-response pair will give us one bandwidth estimate. We could store these measurements in a list and then compute the average over the list at the end. However, by keeping track of the sum and the count, we can get the average at any time by storing only two values.
Storing the values in a list gives us more options though, for example for weighting each bandwidth estimate by file size or duration, so it might be a good idea to switch if we decide to implement that.

theBeardA:
> by keeping track of the sum and the count, we can get the average at any time by storing only two values
I like the idea

> Storing the values in a list gives us more options though, for example for weighting each bandwidth estimate by file size or duration
Do we really need the weightage in this simple calculation? I will leave will up to if you see any benefit from that. I can approve if you are not going with the list approach

theBeardA (approved):

PR #463, opened 2023-03-29
feat(examples): add basic content req/res impl and move some bytes!

ozwaldorf:
## Why

Move some unencrypted bytes with a simple request response impl

## Demo

image

## Checklist

### DIFF
--- a/crates/ursa-pod/examples/client.rs
+++ b/crates/ursa-pod/examples/client.rs
@@ -1,30 +1,33 @@
-use std::error::Error;
-
+use bytes::BytesMut;
 use futures::SinkExt;
 use tokio::net::{TcpStream, ToSocketAddrs};
 use tokio_stream::StreamExt;
 use tokio_util::codec::Framed;
-use tracing::{info, Level};
+use tracing::{debug, info, Level};
 use tracing_subscriber::FmtSubscriber;
-use ursa_pod::codec::{UrsaCodec, UrsaFrame};
+use ursa_pod::codec::{Blake3CID, UrsaCodec, UrsaCodecError, UrsaFrame};
 
 #[tokio::main]
-async fn main() -> Result<(), Box<dyn Error>> {
+async fn main() -> Result<(), UrsaCodecError> {
     let subscriber = FmtSubscriber::builder()
         .with_max_level(Level::TRACE)
         .finish();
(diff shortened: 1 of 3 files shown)


crates/ursa-pod/src/client.rs
```diff
@@ -0,0 +1,85 @@
...
+                }
+
+                debug!("streaming content ({content_len})");
+                if content_len != 0 {
+                    self.transport
+                        .codec_mut()
+                        .read_buffer(content_len as usize, 16384);
+                    match self.transport.next().await.expect("content buffer")? {
+                        UrsaFrame::Buffer(bytes) => Ok(bytes),
+                        f => Err(UrsaCodecError::UnexpectedFrame(f.tag().unwrap())),
```
qti3e, line 79: Handle the complete read of the response when there are more than 16KiB in the response. I think the `UfdpResponse` should be its own struct implementing `AsyncRead` that reads the content up to completion and handles any verification state.

Although right now you can just return the response bytes as they come in for this iteration, in the future, our intended behavior is for the `UfdpResponse` to only output validated chunks of data.

```rs
impl UfdpClient {
  fn request(&mut self, cid: ...) -> UrsaResponse {
    ..
  }
}

struct UrsaResponse<'client> {
  client: &mut 'client,
  bytes: Bytes
}

impl AsyncRead for UrsaResponse {
..
}

**Review by qti3e — CHANGES_REQUESTED** [review]
Looks good and overall in the right direction; I added some comments.

File: crates/ursa-pod/src/client.rs (hunk 1 above)
Comment (ozwaldorf) [reply crates/ursa-pod/src/client.rs]: implemented `Stream` instead, which can be wrapped with `tokio_util::io::StreamReader` for `AsyncRead`

File: crates/ursa-pod/src/client.rs [hunk 2]
```diff
@@ -0,0 +1,158 @@
...
+                    Ok(UfdpResponse {
+                        client: self,
+                        // todo: proof
+                        current_block: BytesMut::with_capacity(256 * 1024),
+                        index: 0,
+                        content_len,
+                    })
+                } else {
+                    Err(UrsaCodecError::Unknown)
+                }
```
qti3e, line 153: move this logic to `UfdpResponse::new().await?`, but could be done later too, because we want the UfdpResponse to handle the proof.

PR #466, opened 2023-03-30
feat: latency measurements

matthias-wright:
## Why
We need measurements for the metrics that will be used for the reputation system, see https://github.com/fleek-network/ursa/issues/452. This PR addresses latency measurements.

## What
- Extended `MeasurementManager` so that we can register RTTs for each peer.
- Latency estimates will be computed from RTTs, assuming a symmetric latency between peers (see diagram below).

## Notes
latency_estimate

* For the diagram above, the latency will be estimated as follows:
`latency = (t4 - t1) / 2`
* We have to make the assumption that `t3 - t2` is relatively small.

## Checklist

### DIFF
--- a/crates/ursa-network/src/measurements/bandwidth.rs
+++ b/crates/ursa-network/src/measurements/bandwidth.rs
@@ -1,0 +1,66 @@
+use std::collections::HashMap;
+use std::time::{Duration, Instant};
+
+pub type Bytes = u128;
+pub type BytesPerSecond = f64;
+pub type RequestId = String;
+
+#[derive(Clone)]
+pub struct Bandwidth {
+    requests: HashMap<RequestId, Request>,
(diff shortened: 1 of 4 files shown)


crates/ursa-network/src/measurements/bandwidth.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub type Bytes = u128;
+pub type BytesPerSecond = f64;
+pub type RequestId = String;
+
+#[derive(Clone)]
+pub struct Bandwidth {
+    requests: HashMap<RequestId, Request>,
+    sum: BytesPerSecond,
+    count: u64,
```
kckeiks, line 12: What is this counting?

kckeiks (approved):
LGTM

PR #492, opened 2023-04-09
Pod/tree

qti3e:
## Why

This implements basic functionality for verifiable streams using Blake3.

## What

Some basic data structures for generating proofs for content and pruning a tree, and validating them incrementally.

## Checklist

- [-] I have run the app using my feature and ensured that no functionality is broken

### DIFF
--- a/crates/ursa-pod/src/tree.rs
+++ b/crates/ursa-pod/src/tree.rs
@@ -48,11 +48,22 @@ impl IncrementalVerifier {
 
         let hash = block.finalize(self.is_root());
         assert_eq!(&hash, self.current_hash());
-        self.next();
+
+        // If we're in the root node (when the data is one block or less), we
+        // should not move to the next.
+        if !self.is_root() {
+            self.next();
+        }
     }
 
     /// Go to the next element in the tree.
+    ///
+    /// # Assumes
+    ///
+    /// This function assumes that is not called on the `root` node.
     fn next(&mut self) {
+        debug_assert!(!self.is_root(), "Next should not be called on the root.");
+
         // To traverse to the next node in the tree we need to follow the
         // following algorithm:
         //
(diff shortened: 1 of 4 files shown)


crates/ursa-pod/fuzz/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "ursa-pod-fuzz"
```
ozwaldorf, line 2: can we add this to `ursa-pod` Cargo.toml workspace?

qti3e: no, this is a standard `cargo fuzz` subproject created by `cargo fuzz init`, checkout other repos.

crates/ursa-pod/src/tree.rs
```diff
@@ -0,0 +1,1622 @@
...
+        if self.is_done() {
+            return Err(IncrementalVerifierError::VerifierTerminated);
+        }
+
+        let hash = block.finalize(self.is_root());
+        self.verify_hash(&hash)
+    }
+
+    /// Go to the next element in the tree.
+    fn move_to_next(&mut self) {
```
b0xtch, line 142: This function scares me 🙃

b0xtch (approved):
Mint

There is generally a large usage of `unsafe` but as we discussed necessary while we push the limits. I do appreciate the fact that you have also added all the tests AND fuzzing which could expose some black holes. Left some completely ignorable comments truck along

crates/ursa-pod/src/tree.rs
```diff
@@ -0,0 +1,1622 @@
...
+    /// Return the index of the target element in the array representation of the
+    /// complete tree.
+    pub fn tree_index(&self) -> usize {
+        self.target
+    }
+}
+
+/// The position of a element in an element in a binary tree.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Direction {
```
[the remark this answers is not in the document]
qti3e: Not when you also have the `Target` case; the direction will translate into one bit anyway at the end, and using a bool here will not give much performance benefit compared to the explicitness of the walker.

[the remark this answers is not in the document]
b0xtch: Makes sense

crates/ursa-pod/src/tree.rs
b0xtch: 😄😅😃🙂🙃😐😶

PR #498, opened 2023-04-11
chore: update PR based triggers for gh workflows to skip draft PRs

jsonsivar:
## Why

This is was mentioned on Discord by @qti3e so I thought I'd make a small PR to get familiar with the CI setup and add that filter.  Not sure if this works yet but will create the draft PRs and see.

Fixes # (issue)

## What

- Each workflow file will now have a setting on the PR based triggers to only trigger on PRs that are ready for review

## Notes

See note on the Terraform build - feel like we can maybe only run those on merges to main?

## Checklist

### DIFF
--- a/.github/workflows/_docker-publish.yml
+++ b/.github/workflows/_docker-publish.yml
@@ -7,6 +7,8 @@ on:
       - "Cargo.toml"
       - ".github/workflows/rust.yml"
       - ".github/workflows/docker-publish.yml"
+    types: [review_requested, ready_for_review]
+
 jobs:
   build:
     name: Docker build and publish ${{ matrix.make.name }} (${{ matrix.os }})
--- a/.github/workflows/_docker-publish.yml
+++ b/.github/workflows/_docker-publish.yml
@@ -7,10 +7,10 @@ on:
       - "Cargo.toml"
       - ".github/workflows/rust.yml"
       - ".github/workflows/docker-publish.yml"
-    types: [ready_for_review]
 
 jobs:
   build:
+    if: '! github.event.pull_request.draft'
     name: Docker build and publish ${{ matrix.make.name }} (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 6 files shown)


.github/workflows/_docker-publish.yml
```diff
@@ -7,6 +7,8 @@ on:
       - "Cargo.toml"
       - ".github/workflows/rust.yml"
       - ".github/workflows/docker-publish.yml"
+    types: [review_requested, ready_for_review]
```
ozwaldorf, line 10: Should we make it only ready for review, so that a draft where some are requested to look doesn't get run before marking ready?

jsonsivar: Yeah that's a good point. Updating.

.github/workflows/terraform.yml
```diff
@@ -6,9 +6,13 @@ on:
       - main
     paths:
       - 'infra/tf-next/**'
+  # question: do we need the PR trigger for terraform? maybe just
+  # config check for PRs and the terraform apply only runs for
+  # merges to main?
```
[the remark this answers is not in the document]
jsonsivar: Oh I thought from this it would run on both PRs and merges to main. Like every commit in a PR will run this because we have the trigger on both the push and the PR (I think):
```
on:
  push:
    branches:
      - main
    paths:
      - 'infra/tf-next/**'
  pull_request:
    paths:
      - 'infra/tf-next/**'
    types: [ready_for_review]
```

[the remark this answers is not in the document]
ozwaldorf: https://github.com/fleek-network/ursa/blob/cc6566dfd32d60b0fca257ea9c6725cd62d721d3/.github/workflows/terraform.yml#L87
This line prevents the actual apply from running unless it's on main

.github/workflows/_docker-publish.yml
ozwaldorf: seems like it's not how we expected, prob should revert

b0xtch (approved):
image

ozwaldorf:
I think all we need to do now is restore the `if` checks to skip, since the CI will get correctly triggered now when marking ready_for_review

PR #507, opened 2023-04-17
fix: stream response completely when skipping cache

kckeiks:
## Why

When file is too large to cache, current algorithm stops streaming response.

## What

- Stream response even when the file is too large to cache.

## Checklist

### DIFF
--- a/crates/ursa-proxy/src/core/handler.rs
+++ b/crates/ursa-proxy/src/core/handler.rs
@@ -181,16 +181,17 @@ pub async fn proxy_pass<C: Cache>(
                 );
                 let stream_body_fut = async move {
                     let mut bytes = Vec::new();
+                    let mut skip_cache = false;
                     while let Some(buf) = body.data().await {
                         match buf {
                             Ok(buf) => {
                                 if let Err(e) = writer.write_all(buf.as_ref()).await {
                                     warn!("Failed to write to stream for {e:?}");
                                 }
-                                bytes.put(buf);
-                                if max_size_cache_entry > 0 && bytes.len() > max_size_cache_entry {
-                                    debug!("Data exceeds max size for a cache entry");
-                                    return;
+                                if !skip_cache {
+                                    bytes.put(buf);
+                                    skip_cache = max_size_cache_entry > 0
+                                        && bytes.len() > max_size_cache_entry;
                                 }
                             }
                             Err(e) => {
@@ -198,6 +199,10 @@ pub async fn proxy_pass<C: Cache>(
                                 return;
                             }
                         }
+                    }
+                    if skip_cache {
+                        debug!("Data exceeds max size for a cache entry");
+                        return;
                     }
                     cache_client.insert(path, bytes);
                 };


crates/ursa-proxy/src/core/handler.rs
```diff
@@ -199,6 +200,10 @@ pub async fn proxy_pass<C: Cache>(
                             }
                         }
                     }
+                    if skip_cache {
```
ozwaldorf, line 203: Related to comment above
```suggestion
                    if max_size_cache_entry > 0 && bytes.len() > max_size_cache_entry {
```

crates/ursa-proxy/src/core/handler.rs
```diff
@@ -181,16 +181,17 @@ pub async fn proxy_pass<C: Cache>(
...
                                     warn!("Failed to write to stream for {e:?}");
                                 }
-                                bytes.put(buf);
-                                if max_size_cache_entry > 0 && bytes.len() > max_size_cache_entry {
-                                    debug!("Data exceeds max size for a cache entry");
-                                    return;
+                                if !skip_cache {
+                                    bytes.put(buf);
+                                    skip_cache = max_size_cache_entry > 0
+                                        && bytes.len() > max_size_cache_entry;
```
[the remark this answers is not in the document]
ozwaldorf: Nevermind, realized you want to skip appending to the byte vec as soon as the content is past the limit

ozwaldorf (approved):
dank

crates/ursa-proxy/src/core/handler.rs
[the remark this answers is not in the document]
kckeiks: The point here is to avoid loading data in memory (in `buf`).

matthias-wright (approved):

b0xtch (approved):

crates/ursa-proxy/src/core/handler.rs
ozwaldorf, line 194: this could be if instead of assigning

kckeiks: You'd still be doing this check on every run until skip_cache is true. This would be more code.

ozwaldorf: I know it's a micro-optimization, but while it may be more rust code, it results in better assembly and less work for the cpu, especially since this part of the code is run many many times. With the if it can do comparison on the stack and immediately JA (jump if comparison is true) to a single instruction to assign skip cache true. Vs the current way which needs to store the value into a register and load that into the heap. 

Here's a extremely simplified compiler explorer example: https://godbolt.org/z/4h9r7b4KM

Direct assignment each iteration:
```asm
.LBB31_10:
  mov rax, qword ptr [rsp]
  mov rcx, qword ptr [rsp + 24]
  cmp rax, rcx // comparison
  seta al // load to register
  and al, 1
  mov byte ptr [rsp + 79], al // assign from the register
  jmp .LBB31_4 // next iteration
```

If conditional and assign if true:
```asm
.LBB31_10:
  mov rax, qword ptr [rsp]
  mov rcx, qword ptr [rsp + 24]
  cmp rax, rcx // comparison len vs max_len
  ja .LBB31_13 // jump if comparison is true
  jmp .LBB31_4 // otherwise, next iteration

.LBB31_13:
  mov byte ptr [rsp + 79], 1 // static assignment to true
  jmp .LBB40_4 // next iteration
```

kckeiks: What is the performance gain in time?

PR #516, opened 2023-04-20
pod: add QUIC benchmarks

### LINKED ISSUE #489: test: write benchmark for UFDP + QUIC

kckeiks:
## Why

Task of #489 

- I removed the feature flags because AFAIK the plan is to run all tests by default. Please let me know if that's not right.
- Client/Server are using a user-provided stream, I added some code to have them return the stream so that users can clean up the stream as needed by the protocol in which UFDP in running on. Open to suggestions of course.

## Checklist

### DIFF
--- a/crates/ursa-pod/Cargo.toml
+++ b/crates/ursa-pod/Cargo.toml
@@ -27,6 +27,11 @@ tracing = { version = "0.1", optional = true }
 # Criterion benchmark dependencies
 hyper = { version = "1.0.0-rc.3", features = ["full"], optional = true }
 http-body-util = { version = "0.1.0-rc.2", optional = true }
+quinn = { version = "0.9.3", optional = true }
+rcgen = { version = "0.10.0", optional = true }
+rustls = { version = "0.20.8", default-features = false, features = ["dangerous_configuration"], optional = true }
+s2n-quic = { version = "1.18.0", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"] , optional = true }
+zeroize_derive = { version = "<1.4", optional = true }
 
 # Binary benchmark dependencies
 serde = { version = "1.0", features = ["derive"], optional = true }
@@ -47,6 +52,8 @@ client = ["tokio", "tokio-stream", "futures", "tracing"]
 server = ["tokio", "tokio-stream", "futures", "tracing"]
 bench-hyper = ["tokio", "hyper", "http-body-util"]
 benchmarks = ["serde", "serde_json", "rayon", "fnv", "clap", "gnuplot"]
+bench-quinn = ["tokio", "quinn", "rcgen", "rustls"]
+bench-s2n-quic = ["tokio", "s2n-quic", "zeroize_derive", "rcgen", "rustls"]
 
 [[bench]]
 name = "encrypt"
--- a/crates/ursa-pod/Cargo.toml
+++ b/crates/ursa-pod/Cargo.toml
@@ -27,10 +27,10 @@ tracing = { version = "0.1", optional = true }
 # Criterion benchmark dependencies
 hyper = { version = "1.0.0-rc.3", features = ["full"], optional = true }
 http-body-util = { version = "0.1.0-rc.2", optional = true }
-quinn = { version = "0.9.3", optional = true }
-rcgen = { version = "0.10.0", optional = true }
-rustls = { version = "0.20.8", default-features = false, features = ["dangerous_configuration"], optional = true }
-s2n-quic = { version = "1.18.0", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"] , optional = true }
+quinn = { version = "0.9", optional = true }
+rcgen = { version = "0.10", optional = true }
+rustls = { version = "0.20", default-features = false, features = ["dangerous_configuration"], optional = true }
+s2n-quic = { version = "1.18", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"] , optional = true }
 
 # Binary benchmark dependencies
 serde = { version = "1.0", features = ["derive"], optional = true }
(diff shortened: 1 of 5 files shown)


crates/ursa-pod/Cargo.toml
```diff
@@ -27,6 +27,12 @@ tracing = { version = "0.1", optional = true }
 # Criterion benchmark dependencies
 hyper = { version = "1.0.0-rc.3", features = ["full"], optional = true }
 http-body-util = { version = "0.1.0-rc.2", optional = true }
+quinn = { version = "0.9.3" }
+rcgen = { version = "0.10.0" }
+rustls = { version = "0.20.8", default-features = false, features = ["dangerous_configuration"] }
+s2n-quic = { version = "1.18.0", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"] }
+# TODO: Add link to issue explaining this.
+zeroize_derive = { version = "<1.4" }
```
ozwaldorf, line 35: I think we still want a feature flag `bench-quic` for these and to make the dependencies optional. Atm for example, a user of the library would end up pulling in/compiling all these deps just to use client/server on their own transport. In terms of e2e.rs, it'd be cfg under the flag but additive instead of replacing the entire benchmark. 

For example, with that change I could run all benchmarks with:
```
cargo bench --bench e2e --features bench-hyper --features bench-quic
```

crates/ursa-pod/src/client.rs
```diff
@@ -143,6 +143,10 @@ where
         Ok(size)
     }
 
+    pub fn finish(self) -> S {
```
ozwaldorf, line 146: Small one, can you just add a doc comment here?

crates/ursa-pod/benches/e2e.rs
```diff
@@ -267,5 +292,282 @@ mod http_hyper {
...
+            });
+        }
+    }
+
+    pub async fn client_loop(addr: String, iterations: usize) {
+        let mut tasks = vec![];
+        let mut endpoint = Endpoint::client("0.0.0.0:0".parse().unwrap()).unwrap();
+        let client_config = quinn::ClientConfig::new(Arc::new(client_config()));
+        endpoint.set_default_client_config(client_config);
+        let stream = endpoint
```
kckeiks, line 364: @qti3e @ozwaldorf what's nice about using QUIC here is that we don't have to open `iterations` connections like in TCP (if we were using TCP + TLS that would be a lot of connection setup overhead). Instead we can open one connection and send each request in a separate stream. They all get multiplexed in the wire.

crates/ursa-pod/benches/e2e.rs
```diff
@@ -114,55 +118,76 @@ fn protocol_benchmarks(c: &mut Criterion) {
                 http_hyper::server_loop,
             );
         }
+
+        #[cfg(feature = "bench-quic")]
+        {
+            let mut g = c.benchmark_group(format!("QUINN UFDP/{range}"));
```
ozwaldorf, line 124: minor nit, could you add `QUIC` so that both implementations can be run at once?

ozwaldorf, line 124: Minor nit so both impl can be run with `QUIC`
```suggestion
            let mut g = c.benchmark_group(format!("QUINN QUIC UFDP/{range}"));
```

crates/ursa-pod/benches/e2e.rs
```diff
@@ -267,5 +292,282 @@ mod http_hyper {
...
+        }
+        join_all(tasks).await;
+    }
+}
+
+#[cfg(feature = "bench-quic")]
+mod tls_utils {
+    use std::sync::Arc;
+
+    pub struct SkipServerVerification;
```
kckeiks, line 527: I forgot, I'm going to remove this because it would be better to not skip the certificate verification so we can get a better idea of the performance.

crates/ursa-pod/benches/e2e.rs
ozwaldorf: This is a really great point, and good to consider when we choose the main mode of transport!

crates/ursa-pod/benches/e2e.rs
ozwaldorf: Since the iterations in the bench are multiplexed over a single quic connection, this would only affect the initial connection right?

kckeiks: Yes, that's right.

crates/ursa-pod/benches/e2e.rs
```diff
@@ -267,5 +292,282 @@ mod http_hyper {
...
+
+    const CLIENT_PUB_KEY: [u8; 48] = [3u8; 48];
+    const CID: Blake3Cid = Blake3Cid([3u8; 32]);
+
+    pub async fn server_loop(
+        addr: String,
+        content: &'static [u8],
+        tx_started: tokio::sync::oneshot::Sender<u16>,
+    ) {
+        let server_config = ServerConfig::with_crypto(Arc::new(server_config()));
```
kckeiks, line 321: TODO: Make TLS configurable.

ozwaldorf:
Just for visibility, s2n-quic is having an issue on ubuntu 22.04:
```
Benchmarking S2N-QUIC UFDP/Content Size (Kilobyte)/1 request/1: Collecting 20 samples in estimated 11.116 s (8190 iterations)
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: MaxHandshakeDurationExceeded { max_handshake_duration: 10s, source: Location { file: "/home/oz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/s2n-quic-transport-0.18.2/src/connection/connection_imp
l.rs", line: 1027, col: 24 } }', crates/ursa-pod/benches/e2e.rs:510:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: bench failed, to rerun pass `-p ursa-pod --bench e2e`
```

kckeiks:
Closing for #519

PR #519, opened 2023-04-21
pod: add QUIC and TCP/TLS e2e benchmarks

kckeiks:
## Why

Task of #489 

## Checklist


crates/ursa-pod/Cargo.toml
```diff
@@ -42,11 +45,13 @@ criterion = { version = "0.4.0", features = ["html_reports", "async_tokio"] }
...
 
 [features]
-default = ["client", "server"]
+default = ["client", "server", "tls"]
 client = ["tokio", "tokio-stream", "futures", "tracing"]
 server = ["tokio", "tokio-stream", "futures", "tracing"]
 bench-hyper = ["tokio", "hyper", "http-body-util"]
 benchmarks = ["serde", "serde_json", "rayon", "fnv", "clap", "gnuplot"]
+bench-quic = ["quinn", "tls"]
+tls = ["rustls", "rcgen"]
```
ozwaldorf, line 54: oops!

ozwaldorf (approved):
Looking good overall, although it's strange that we seem to hit a cap at 360MiB/s. We'll have a better comparison with tls+tcp once that's in

kckeiks:
> Looking good overall, although it's strange that we seem to hit a cap at 360MiB/s. We'll have a better comparison with tls+tcp once that's in

We haven't even started tweaking the parameters yet so I wouldn't worry atm. The point of these benches is to test and measure before committing to anything.

crates/ursa-pod/Cargo.toml
kckeiks: ?

ozwaldorf (approved):
dank! 

The only minor comment is I'd be curious to see how quinn performs when creating unique concurrent connections vs multiplexed like it is now. I know it's technically "against" how quic should be used, but if it's something in the implementation of quinn limiting the multiplexed throughput, and separate streams are faster, we should use it in the way that moves the most bytes in the shortest time. If the implementation improves, it seems to me like we could switch back to multiplex later on with minimal work.

kckeiks:
> dank!
> 
> The only minor comment is I'd be curious to see how quinn performs when creating unique concurrent connections vs multiplexed like it is now. I know it's technically "against" how quic should be used, but if it's something in the implementation of quinn limiting the multiplexed throughput, and separate streams are faster, we should use it in the way that moves the most bytes in the shortest time. If the implementation improves, it seems to me like we could switch back to multiplex later on with minimal work.

Imo it's not a good idea for a client to open multiple connections when requesting a set of files. When using the network, you have to consider the network's resources. If a user needs to download 5 files, these can be done in 5 streams in one connection. What you're suggesting is opening 5 different connections which means 5 separate handshakes, more RTTs, more network activity and thus more network resources being used on both client and server. This is inefficient and may lead to network congestion issues.

kckeiks:
> > dank!
> > The only minor comment is I'd be curious to see how quinn performs when creating unique concurrent connections vs multiplexed like it is now. I know it's technically "against" how quic should be used, but if it's something in the implementation of quinn limiting the multiplexed throughput, and separate streams are faster, we should use it in the way that moves the most bytes in the shortest time. If the implementation improves, it seems to me like we could switch back to multiplex later on with minimal work.
> 
> Imo it's not a good idea for a client to open multiple connections when requesting a set of files. When using the network, you have to consider the network's resources. If a user needs to download 5 files, these can be done in 5 streams in one connection. What you're suggesting is opening 5 different connections which means 5 separate handshakes, more RTTs, more network activity and thus more network resources being used on both client and server. This is inefficient and may lead to network congestion issues.

I'm mostly concerned on opening an unbounded number of connections per client. If we decide to use multiple connections on our clients, let's put a cap on the number of independent connections that a client can make. Let's make a good trade off.

ozwaldorf:
We do have a protocol level limit for one node to a single paying client, established by the `lanes` with an upper bound of 254, but we will probably use a much lower number, initially proposed to be 24

PR #520, opened 2023-04-22
pod: frame filters and alloc improvement

ozwaldorf:
## Why

resolves todo for implementing frame filters, and improves some allocations for the inner buffer.

## What

- add frame filter impl
- bring back read_buffer and reserve the len in the inner buffer
- whenever reading a frame, reserve the len again to make sure self.buffer always has at least 148 bytes available (max frame size)
- also added some minor cleanup in example server and ufdp-bench-parser
- improved code comments and added some todos

## Checklist

### DIFF
--- a/crates/ursa-pod/src/server.rs
+++ b/crates/ursa-pod/src/server.rs
@@ -2,12 +2,13 @@ use std::sync::Arc;
 
 use bytes::BytesMut;
 use tokio::io::{AsyncRead, AsyncWrite};
-use tracing::error;
 
 use crate::{
     connection::{
-        consts::{CONTENT_REQ_TAG, HANDSHAKE_REQ_TAG},
-        Reason, UfdpConnection, UrsaCodecError, UrsaFrame,
+        consts::{
+            CONTENT_RANGE_REQ_TAG, CONTENT_REQ_TAG, DECRYPTION_KEY_REQ_TAG, HANDSHAKE_REQ_TAG,
+        },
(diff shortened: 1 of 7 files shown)


kckeiks (commented):
LGTM.

Nit:
`// Doc and comment consistency, preferably capitalize the first word and end should be marked with a period.`

crates/ursa-pod/src/server.rs
```diff
@@ -180,26 +176,21 @@ impl<S: AsyncWrite + AsyncRead + Unpin, B: Backend> UfdpHandler<S, B> {
...
                 }
-                Some(f) => {
-                    error!("Terminating asdf, unexpected frame: {f:?}");
-                    self.conn
-                        .termination_signal(Some(Reason::UnexpectedFrame))
-                        .await?;
-                    return Err(UrsaCodecError::UnexpectedFrame(f.tag().unwrap()));
-                }
                 None => return Err(UrsaCodecError::Unknown),
+                Some(_) => unreachable!(),
```
qti3e, line 189: Don't trust the client to oblige to the protocol. we should never panic here.

qti3e (changes requested):

crates/ursa-pod/src/server.rs
ozwaldorf: the frame filter should guarantee unreachable there

qti3e (approved):
eyval.

PR #526, opened 2023-04-25
Refactor: Unix domain sockets for abci communication

daltoncoder:
## Why

Switches from TCP to unix sockets for faster communication between application and consensus. The ABCI client we were using before was synchronous and blocking so created async non blocking client instead. 

We dont want rpc requests blocking the ABCI query client, and we want to keep the client connection alive for fast innerprocess querys so ive also added all rpc querys will spawn a new task with its own client and the query client we keep alive will only be used for innerproccess querys

## What

- switched to unix socket abci communication
- all rpc querys will run in their own task with their own client

## Checklist


crates/ursa-application/Cargo.toml
```diff
@@ -22,13 +22,18 @@ tracing-subscriber.workspace = true
...
+
+revm = { version = "3.1.0", features = ["serde"] }
 hex = "0.4.3"
 tendermint-abci = { version = "0.23.7", features = ["client"] }
 tendermint-proto = "0.23.7"
 reqwest = "0.11.11"
+parking_lot = "0.12.1"
 ursa-utils = { path = "../ursa-utils" }
-
+prost = { version = "0.10.4", default-features = false, features = [
```
kckeiks, line 37: `prost` requires us to install protoc (C++) and it would make our software lighter if we could remove it. Could we please consider using a pure rust alternative like [`quick-protobuf`](https://docs.rs/quick-protobuf/latest/quick_protobuf/)? For the most part it performs better than `prost` https://github.com/tafia/quick-protobuf/tree/master/perftest#results.

daltoncoder: Im going to look into this more. Im actually only bringing in prost for its message trait here, since the ABCI server crate uses prost. We are already depend on Prost in a lot of our our crates(like Narwhal, Libp2p) so im not  sure it would help with being lighter. But I could be down for making our own ABCI server implementation if it is significant performance increase

matthias-wright (approved):
lgtm

PR #527, opened 2023-04-26
feat: blake3 integration for client and server

ozwaldorf:
## Why

Integrates blake3 verified streaming to client and server implementations

## What

- Replace `types::Blake3Cid` with `blake3::Hash`
- Add `Backend::get_tree()`
- (server) send proofs
- (client) receive proofs and verify content

## Checklist

### DIFF
--- a/crates/ursa-pod/src/client.rs
+++ b/crates/ursa-pod/src/client.rs
@@ -1,8 +1,9 @@
+use std::io::{Error, ErrorKind};
+
+use blake3::{ursa::BlockHasher, Hash};
 use bytes::{BufMut, BytesMut};
 
 use tokio::io::{AsyncRead, AsyncWrite};
-
-use tracing::debug;
 
 use crate::{
     connection::{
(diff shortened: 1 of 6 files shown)


crates/ursa-pod/src/client.rs
```diff
@@ -77,24 +82,31 @@ where
...
-                                if proof_buf.len() == len {
-                                    // todo: decode proof
-                                    break;
+                    if proof_len != 0 {
+                        let len = proof_len as usize;
+                        self.conn.read_buffer(len);
+                        let mut proof_buf = BytesMut::with_capacity(len);
+                        loop {
+                            match instrument!(
+                                self.conn.read_frame(None).await?,
```
qti3e, line 91: isn't read_frame supposed to return the frame when the data len species by `read_buffer(len)` is provided? is this loop necessary?

qti3e (dismissed):
looks good, just left one comment

crates/ursa-pod/src/client.rs
ozwaldorf: I think this makes sense, I can refactor it

